PR #23, opened 2015-08-05
Don't abuse /lib/systemd, remove init scripts on package removal

fpytloun:
Automatically generated systemd startup files should not go to /lib but /etc
Best approach would be to deliver this in a normal way by package itself - only then it should go to /lib/systemd.

### DIFF
--- a/ovs/extensions/services/systemd.py
+++ b/ovs/extensions/services/systemd.py
@@ -30,7 +30,7 @@ class Systemd(object):
     @staticmethod
     def _service_exists(name, client, path):
         if path is None:
-            path = '/lib/systemd/system/'
+            path = '/etc/systemd/system/'
         file_to_check = '{0}{1}.service'.format(path, name)
         return client.file_exists(file_to_check)
 
@@ -77,9 +77,9 @@ class Systemd(object):
             template_file = template_file.replace('<ADDITIONAL_DEPENDENCIES>', dependencies)
 
         if target_name is None:
-            client.file_write('/lib/systemd/system/{0}.service'.format(name), template_file)
+            client.file_write('/etc/systemd/system/{0}.service'.format(name), template_file)
         else:
-            client.file_write('/lib/systemd/system/{0}.service'.format(target_name), template_file)
+            client.file_write('/etc/systemd/system/{0}.service'.format(target_name), template_file)
             name = target_name
 
         try:
@@ -104,7 +104,7 @@ class Systemd(object):
     def remove_service(name, client):
         # remove systemd.service file
         name = Systemd._get_name(name, client)
-        client.file_delete('/lib/systemd/system/{0}.service'.format(name))
+        client.file_delete('/etc/systemd/system/{0}.service'.format(name))
         client.run('systemctl daemon-reload')
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


jamie-liu:
Use /lib/systemd/system/ is because the systemd in Ubuntu 15.04 doesn't put the service files under path /usr/lib/systemd/system/, while Centos does.

fpytloun:
I have set up Ubuntu Vivid and it uses /etc/systemd as expected..

```
root@vivid:/etc/systemd/system# cat << EOF > test.service
> [Unit]
> Description=Test
> 
> [Service]
> ExecStart=/bin/bash -c 'echo test'
> EOF
root@vivid:/etc/systemd/system# systemctl start test.service
root@vivid:/etc/systemd/system# systemctl status test.service
● test.service - Test
   Loaded: loaded (/etc/systemd/system/test.service; static; vendor preset: enabled)
   Active: inactive (dead)
```

So it's ok to use it for generated startup files and don't abuse /lib which should contain only package-shipped stuff (as almost all locations except for /var and /etc).
If you have no further objections, I will rebase the pull request.

jamie-liu:
OK, I agree. Thanks.

jamie-liu:
I don't think it's a good idea to use /etc/systemd as the service file path for systemd. 
The Centos 7 is by default using /usr/lib/systemd/system/, while Ubuntu 15.04  is by default using /lib/systemd/system/.
The service files under /etc/systemd/system/ are just symbol links. When doing systemclt enable <.service> the symbol link is created, while doing systemclt disable <.service>  the  symbol link  is removed automatically.

[root@~]#systemctl enable ovs-logstash
ln -s '/usr/lib/systemd/system/ovs-logstash.service' '/etc/systemd/system/multi-user.target.wants/ovs-logstash.service'

[root@~]#ls /etc/systemd/system/multi-user.target.wants/  -l | grep logstash
lrwxrwxrwx  1 root root 44 Aug 24 15:49 ovs-logstash.service -> /usr/lib/systemd/system/ovs-logstash.service

[root@~]#systemctl disable ovs-logstash
rm '/etc/systemd/system/multi-user.target.wants/ovs-logstash.service'

khenderick:
On my own laptop, running ArchLinux, I see a similar behaviour as reported by @jamie-liu; The service files under `/etc/systemd/system/` are symbolic links to `/usr/lib/systemd/system/`.

```
[root@laptop system]# ls -al /etc/systemd/system/multi-user.target.wants/zfs.target
lrwxrwxrwx 1 root root 34 Apr 19  2014 /etc/systemd/system/multi-user.target.wants/zfs.target -> /usr/lib/systemd/system/zfs.target
```

It also creates and removes symlinks when using systemctl to enable and disable services.

fpytloun:
Yes, it's used to make symlinks, but it can be also used for admin's custom startup files (in a similar manner as per-user ~/.config/systemd).

Basic FHS rule is, don't ever touch /usr, /lib and paths like this by hand as these paths are exclusively for **static** files. Generated startup files are not static, so using these paths will only make a useless mess.
This would be never accepted in upstream distribution (Debian nor RHEL) as it's simply breaking the rule and all these distros are FHS-compliant.
Some of my slides on FHS and Debian packaging: https://fpy.cz/pub/slides/debian-packaging/#/step-2

Or the best would be to avoid generating them, deliver static startup files by packages in a standard way. I don't think it should be difficult do to it this way.

jamie-liu:
OK, I agree to put the OVS service file under /etc/systemd/system/.
Just did a quick test to verify it and it works.

[root@system]# ls /etc/systemd/system/test.service  -l
-rw-r--r-- 1 root root 193 Aug 24 17:17 /etc/systemd/system/test.service
[root@system]# 
[root@system]# ls -l /etc/systemd/system/multi-user.target.wants/test.service
ls: cannot access /etc/systemd/system/multi-user.target.wants/test.service: No such file or directory
[root@system]# 
[root@system]# systemctl enable test.service
ln -s '/etc/systemd/system/test.service' '/etc/systemd/system/multi-user.target.wants/test.service'
[root@system]# 
[root@system]# ls -l /etc/systemd/system/multi-user.target.wants/test.service
lrwxrwxrwx 1 root root 32 Aug 24 17:21 /etc/systemd/system/multi-user.target.wants/test.service -> /etc/systemd/system/test.service
[root@system]# 
[root@system]# systemctl disable test.service  
rm '/etc/systemd/system/multi-user.target.wants/test.service'

PR #24, opened 2015-08-05
Use system init scripts

fpytloun:


ovs/extensions/services/upstart.py
```diff
@@ -42,6 +42,8 @@ def _get_name(name, client, path=None):
         """
         if Upstart._service_exists(name, client, path):
             return name
+        if os.path.exists('/etc/init.d/{0}'.format(name)):
```
khenderick: `client.file_exists(...)` should be used

config/templates/systemd/ovs-memcached.service
```diff
@@ -1,13 +0,0 @@
-[Unit]
-Description=memcache
-After=local-fs.target network.target
-
-[Service]
-Type=simple
-ExecStart=/usr/bin/memcached -m 1000 -p 11211 -u memcached -l <MEMCACHE_NODE_IP>
```
khenderick, line 7: Is the default memcached listening on all ips? If not, we need to make sure it at least listens to the public ip selected in `ovs setup`

PR #28, opened 2015-08-09
Allow private key auth for initial setup

fpytloun:
It's no longer needed to have SSH password auth enabled so it's enhancing security.

Needs testing and maybe some fixes on other places.


khenderick:
Replaced by #45

PR #35, opened 2015-08-11
Don't patch PYTHONPATH of cinder-volume anymore

fpytloun:


khenderick:
Checked with @eduardBM wrt review processes. For now, this should be OK and give no issues

PR #37, opened 2015-08-13
Add nvme support and fix bug in systemd manager

jamie-liu:


scripts/install/openvstorage-webapps.postinst.py
```diff
@@ -77,9 +77,9 @@
 
 # Setup nginx site
 if not os.path.exists("/etc/nginx/sites-enabled/openvstorage.conf"):
-    check_output("ln -s /etc/nginx/sites-available/openvstorage.conf /etc/nginx/sites-enabled/openvstorage.conf", shell=True)
+    check_output("ln -sf /etc/nginx/sites-available/openvstorage.conf /etc/nginx/sites-enabled/openvstorage.conf", shell=True)
```
fpytloun: Why to force override of existing files if this code is in `not os.path.exists("/etc/nginx/sites-enabled/openvstorage.conf")` condition?

ovs/extensions/services/systemd.py
```diff
@@ -41,6 +41,8 @@ def _get_name(name, client, path=None):
         """
         if Systemd._service_exists(name, client, path):
             return name
+        if Systemd._service_exists(name, client, '/lib/systemd/system/'):
```
fpytloun: FYI: see this pull request - https://github.com/openvstorage/openvstorage/pull/23

scripts/install/openvstorage-webapps.postinst.py
jamie-liu: OK, will remove it.

PR #48, opened 2015-08-19
OVS-2646 - Handle more scenario's for externally managed clusters

khenderick:


ovs/lib/setup.py
```diff
@@ -1390,18 +1422,24 @@ def _demote_node(cluster_ip, master_ip, cluster_name, ip_client_map, unique_id):
...
+                target_client.run('rabbitmqctl reset; sleep 5;')
+                target_client.run('rabbitmqctl stop; sleep 5;')
+                SetupController._change_service_state(target_client, 'rabbitmq-server', 'stop')
+                target_client.file_unlink("/var/lib/rabbitmq/.erlang.cookie")
 
         print 'Removing services'
         logger.info('Removing services')
-        for service in [s for s in SetupController.master_node_services if s not in (SetupController.extra_node_services + [SetupController.ARAKOON_OVSDB, SetupController.ARAKOON_VOLDRV])]:
+        services = [s for s in SetupController.master_node_services if s not in (SetupController.extra_node_services + [SetupController.ARAKOON_OVSDB, SetupController.ARAKOON_VOLDRV])]
+        if configure_rabbitmq:
```
kvanhijf: Should be 'if not configure_rabbitmq', because if we don't handle rabbitmq and memcache, then we should remove the services, so they are not stopped

PR #51, opened 2015-08-26
OVS-2673 - Fix packaging code

khenderick:


khenderick:
Seems I have to create a new PR anyway.

PR #91, opened 2015-09-08
OVS-2790 - Restored correct volumedriver service name

kvanhijf:


khenderick:
Please use cherry-pick in the future.

PR #110, opened 2015-09-15
OVS-2372 Extra validation on add_vpool task

eduardBM:


eduardBM:
Fixed by latest changes to add_vpool (flexible ssd layout).

PR #118, opened 2015-09-16
OVS-2368 - ArakoonStore.prefix internally uses Arakoon range

khenderick:


ovs/extensions/storage/persistent/arakoonstore.py
```diff
@@ -123,3 +134,19 @@ def _try(method, *args, **kwargs):
...
+        encoding = 'ascii'  # For future python 3 compatibility
+        array = bytearray(str(key), encoding)
+        for index in range(len(array) - 1, -1, -1):
+            array[index] += 1
+            if array[index] < 128:
+                while array[-1] == 0:
+                    array = array[:-1]
+                return str(array.decode(encoding))
+            array[index] = 0
+        return '\xff'
```
domsj, line 152: so you're claiming '\xff' is the biggest possible key?
(btw for this sort of methods unit tests can be handy...)

khenderick, line 152: Yeah, I make a lot of assumptions here on what `key` will be. When converted to a byte array, I do not expect it to have values > 127 (based on the fact that `key` will always be an ascii string). This means that `\xff` is larger than any key a caller will be able to put into Arakoon, and thus can safely be assumed to be the biggest possible key.

PR #122, opened 2015-09-17
OVS-2988 - Clean paramiko client after failed connection attempt

khenderick:


ovs/extensions/generic/sshclient.py
```diff
@@ -154,10 +154,13 @@ def _connect(self):
             self.client.connect(self.ip, username=self.username, password=self.password)
         except socket.error as ex:
             if 'No route to host' in str(ex):
-                message = 'SocketException: No route to host {0}'.format(ip)
+                message = 'SocketException: No route to host {0}'.format(self.ip)
                 logger.error(message)
                 raise UnableToConnectException(message)
             raise
+        except paramiko.AuthenticationException:
```
domsj: Maybe you should do this for all types of exceptions ... after all it seems you don't know what state the client will be in if anything throws here

khenderick:
I see what you did there...

PR #138, opened 2015-09-25
OVS-2214 - Flexible SSD layout

kvanhijf:


khenderick:
Work in progress to track pending changes

PR #139, opened 2015-09-25
Update service files and bug fix

jamie-liu:
1. update service file for Centos 7
2. update dependency for ovs-failovercache in upstart


khenderick:
Fixes internal ticket OVS-3076

PR #157, opened 2015-10-23
OVS-3184 Lock vdisk creation in update_vmachine_config

eduardBM:
- find, init VDisk and save() should be in same locked context

### DIFF
--- a/ovs/lib/vmachine.py
+++ b/ovs/lib/vmachine.py
@@ -598,20 +598,31 @@ class VMachineController(object):
             storagedrivers = StorageDriverList.get_storagedrivers()
             datastores = dict([('{0}:{1}'.format(storagedriver.storage_ip, storagedriver.mountpoint), storagedriver) for storagedriver in storagedrivers])
             vdisk_guids = []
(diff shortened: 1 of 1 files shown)


ovs/lib/vmachine.py
```diff
@@ -598,20 +598,31 @@ def update_vmachine_config(vmachine, vm_object, pmachine=None):
...
+                                vdisk.volume_id = vdisk.storagedriver_client.get_volume_id(str(disk['backingfilename']))
+                                vdisk.size = vdisk.info['volume_size']
+                                # Create the disk in a locked context, but don't execute long running-task in same context
+                                vdisk.save()
+                                ensure_safety = True
+                        finally:
+                            mutex.release()
+                        if ensure_safety:
                             MDSServiceController.ensure_safety(vdisk)
+                        vdisk = VDiskList.get_by_devicename_and_vpool(disk['filename'], datastores[datastore].vpool)
```
kvanhijf: Obsolete since vdisk is retrieved on line 609 (perhaps put it outside of try - except)
vdisk will exist here, so can be removed

PR #162, opened 2015-10-28
RPM upgrade support on Centos 7

jamie-liu:


ovs/extensions/packages/rpm.py
```diff
@@ -41,12 +44,65 @@ def get_versions():
 
     @staticmethod
     def install(package_name, client, force=False):
-        raise NotImplementedError("Installing RPM packages not yet implemented")
+        counter = 0
+        max_counter = 3
+        while True and counter < max_counter:
```
toolslive: `while counter < max_counter:` ?

jamie-liu: yes, I think the true needs to be removed here. Already modified.

khenderick:
Internal reference: OVS-3040

PR #165, opened 2015-10-30
Ovs 3048 ceph librados

eduardBM:


ovs/lib/storagerouter.py
```diff
@@ -997,6 +1070,10 @@ def remove_storagedriver(storagedriver_guid):
         DiskController.sync_with_reality(storagerouter.guid)
 
         # First model cleanup
+        for service in ServiceList.get_services():
```
khenderick, line 1073: Please use the same approach as the alba proxy below.

ovs/lib/storagerouter.py
```diff
@@ -755,6 +798,27 @@ def add_vpool(parameters):
...
+        # Start ovs ceph proxy BEFORE starting volumedriver
+        if vpool.backend_type.code == 'ceph_ovs_proxy':
+            params = {'PORT': str(ceph_proxy_port),
+                      'CEPHCONF': ceph_proxy_config_file,
+                      'VPOOL_NAME': vpool.name}
+            ceph_proxy_service = "ovs-cephproxy_{0}".format(vpool.name)
+            ServiceManager.add_service("ovs-cephproxy", params=params, client=root_client, target_name=ceph_proxy_service)
+            ServiceManager.enable_service(ceph_proxy_service, client=root_client)
+            ServiceManager.start_service(ceph_proxy_service, client=root_client)
+            logger.debug('Starting ceph proxy service {0}'.format(ceph_proxy_service))
```
khenderick, line 810: We don't need to start the ceph service. If configured correctly, it should be a startup dependency of the volumedriver. See alba proxy logic

ovs/lib/storagerouter.py
```diff
@@ -878,6 +948,9 @@ def remove_storagedriver(storagedriver_guid):
         if ServiceManager.has_service(voldrv_service, client=client):
             ServiceManager.disable_service(voldrv_service, client=client)
             ServiceManager.stop_service(voldrv_service, client=client)
+        if ServiceManager.has_service(cephproxy_service, client=client):
+            ServiceManager.disable_service(cephproxy_service, client=client)
+            ServiceManager.stop_service(cephproxy_service, client=client)
         if ServiceManager.has_service(dtl_service, client=client):
             ServiceManager.disable_service(dtl_service, client=client)
             ServiceManager.stop_service(dtl_service, client=client)
```
khenderick, line 956: The below piece of code will have to be extended for the ceph proxy as well. The alba proxy is started below for a short moment of time to allow the volumedriver to clean out the backend. It should be more or less identical.

webapps/frontend/locales/en-US/ovs.json
```diff
@@ -712,6 +712,7 @@
                 "choosebackend": "Choose one of the available backends",
                 "clientid": "Client ID",
                 "clientsecret": "Client secret",
+                "cephfileinvalid": "Invalid Ceph client conf file. Make sure the file was copied from the Ceph Master node to this Storage Router in the correct location and that it points to a valid Ceph Master",
```
khenderick, line 715: We might inform the user about what the correct location is

PR #198, opened 2015-11-19
OVS-3419

eduardBM:


ovs/extensions/services/upstart.py
```diff
@@ -91,18 +92,27 @@ def add_service(name, client, params=None, target_name=None, additional_dependen
             client.file_write('/etc/init/{0}.conf'.format(target_name), template_file)
 
     @staticmethod
-    def get_service_status(name, client):
+    def get_service_status(name, client, return_output=False):
```
toolslive, line 95: The body relies heavily on the `return_output` argument to be a boolean (what if it were `1` ?)
It might be safer to assert the the type first (something like  `assert (type(return_output) is types.BooleanType)` ).
The tests in the body become simpler:
`if return_output: ....`
Also the type of the return value isn't stable. Since you always capture the output, you might want to always return the tuple `(status, output)`. Otherwise, you force a client to change its code if it changes its mind about its interest in the output.

khenderick, line 95: I don't care too much what people pass into the `return_output` argument, if the caller wants output, he should pass in `True`. If they want to pass in their grandma that's fine for me, but it won't give them output.

khenderick, line 95: Well, to be honest; I don't care about the input type enough to justify putting asserts inside the function as in my opinion it makes the code uglier. I'd rather make this type checking decorator we discussed earlier that doesn't mess up the method's body. I'm going to create a ticket for that.

eduardBM, line 95: @toolslive 
Type checking is not "standard" python code, so we don't use it.
Technically, the test "if return_output" has the same result as "if return_output is True" but the second is more restrictive and expressive (as we want a boolean); the first test will pass with any "True-ish" value.
As for the "type" of the return value, that's not important. Making the return_output by default False, means existing clients don't have to change their code when calling these methods, so the impact is minimal; only in places where we explicitly need the output, we pass True and expect a tuple (boolean, string) instead of a string (and also no type-checking or length-checking  is needed as the method is expected to return a string when return_output is False, return a tuple when return_output is True, or raise an exception when it can't return anything).
And if a client changes its mind about the output, they will have to explicitly ask it (pass True to return_output) and be prepared to receive a tuple.

toolslive, line 95: You can find several packages for python2 that have various flavours of a  type decorator.

PR #203, opened 2015-11-25
remove offline node support

jamie-liu:


khenderick:
Merging in "as is". Work in progress.

yongshengma:
It would be great to have a cli subcommand `ovs remove offine node` .  As far as I understand, `ovs remove node <ip>` can only remove a node which services are still available.  It can't process a crashed node or really offline node that can't be accessed any more.

PR #224, opened 2015-11-30
OVS-3317 - Introduced MDS safety of 2 (down from 3)

kvanhijf:


khenderick:
Please repull on eugene

PR #239, opened 2015-12-03
OVS-3659 - Add wizard loading

khenderick:


khenderick:
The solution didn't work as expected after some more testing. Discarding this change and postpone fixing the issue for now.

khenderick:
Partially reverted the changes so we fix the issue but it's not a full solution yet.

PR #249, opened 2015-12-07
OVS-3426 - The DB role should be available on all nodes

khenderick:


webapps/frontend/app/viewmodels/wizards/addvpool/gather_mountpoints.js
```diff
@@ -107,20 +107,12 @@ define([
                     })
                     .done(function() {
                         self.activateResult = { valid: true, reasons: [], fields: [] };
-                        var requiredRoles = ['READ', 'WRITE'];
-                        if (self.data.arakoonFound() === false) {
```
kvanhijf, line 111: is data.arakoonFound still used elsewhere in the code?

khenderick, line 111: Valid point, I'll remove it.

PR #255, opened 2015-12-08
OVS-3705 - Ensure single improvements

kvanhijf:


khenderick:
Branch was based on unstable

PR #273, opened 2015-12-10
Ovs 3755 sticky snapshot

eduardBM:


webapps/frontend/app/viewmodels/site/vdisk-detail.js
```diff
@@ -162,6 +163,12 @@ define([
             self.vDisk().loadParentConfiguration();
             self.loadStorageRouters();
         };
+        self.removeSnapshot = function(snapshotid) {
+            api.post('vdisks/' + self.vDisk().guid() + '/removesnapshot', {
```
khenderick: I'd like to have a confirmation dialog here. There are a bunch of these around the code so you should have something to base on

webapps/frontend/app/views/site/vdisk-detail.html
```diff
@@ -159,6 +159,16 @@ <h4 data-i18n="ovs:vdisks.detail.snapshots"></h4>
                             <td><span data-bind="text: (is_automatic ? 'Automatic' : 'Manual')"></span></td>
                             <td><span class="fa pointer" data-bind="translate: (is_consistent ? 'ovs:icons.consistent' : 'ovs:icons.inconsistent'),
                                                                     tooltip: (is_consistent ? 'ovs:generic.consistentss' : 'ovs:generic.inconsistentss')"></span></td>
+                            <td><span data-bind="text: (is_sticky ? 'ovs:generic.sticky' : '-')"></span></td>
+                            <td>
```
khenderick, line 163: I don't see a column for this td containing the remove button.

ovs/lib/scheduledtask.py
```diff
@@ -191,6 +196,8 @@ def make_timestamp(offset):
         for bucket_chain in bucket_chains:
             for bucket in bucket_chain:
                 for snapshot in bucket['snapshots']:
+                    if snapshot.get('is_sticky') is True:
+                        continue
```
khenderick: There shouldn't be any sticky snapshots in here, based on the filters above, right?

PR #274, opened 2015-12-10
OVS-3671 - archive and cleanup of existing arakoon data

pploegaert:


ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -183,15 +206,23 @@ def create_cluster(cluster_name, ip, exclude_ports, base_dir, plugins=None):
         ports = System.get_free_ports(port_range, exclude_ports, 2, client)
         node_name = System.get_my_machine_id(client)
 
+        home_dir = ArakoonInstaller.ARAKOON_HOME_DIR.format(base_dir, cluster_name)
+        log_dir = ArakoonInstaller.ARAKOON_LOG_DIR.format(cluster_name)
+        tlog_dir = ArakoonInstaller.ARAKOON_TLOG_DIR.format(base_dir, cluster_name)
+
+        ArakoonInstaller.archive_existing_arakoon_data(ip, home_dir, base_dir + '/arakoon/', cluster_name)
+        ArakoonInstaller.archive_existing_arakoon_data(ip, log_dir, '/var/log/arakoon/', cluster_name)
```
khenderick: I'd use sth like `ArakoonInstaller.ARAKOON_LOG_DIR.format('')` here to allow us to change that path by just updating the global variable

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -172,6 +172,29 @@ def __init__(self):
...
+        logger.debug('archive - check if {0} exists'.format(directory))
+        if new_client.dir_exists(directory):
+            logger.debug('archive - from {0}'.format(directory))
+            archive_dir = top_dir + 'archive/' + cluster_name
+            if new_client.dir_exists(archive_dir + '/' + os.path.basename(directory)):
+                logger.debug('archive - from existing archive {0}'.format(archive_dir))
+                timestamp = time.strftime('%Y-%m-%dT%H:%M:%S', time.gmtime())
+                new_archive_dir = archive_dir + '-' + timestamp
+                new_client.dir_create(new_archive_dir)
+                new_client.run('/bin/mv {0} {1}'.format(archive_dir, new_archive_dir))
```
khenderick: It's tricky to use `/bin/mv` as paths tend to change between Ubuntu and CentOS. I think just `mv` is better here

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -271,7 +311,13 @@ def _destroy_node(config, node):
         ArakoonInstaller.remove(config.cluster_id, client=root_client)
 
         # Cleans all directories on a given node
-        root_client.dir_delete([node.log_dir, node.tlog_dir, node.home])
+        for directory in [node.log_dir, node.tlog_dir, node.home]:
+            if root_client.dir_exists(directory):
+                root_client.dir_delete([directory])
```
khenderick: Why is that? The dir_delete already checks whether the directory exists

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -271,7 +311,13 @@ def _destroy_node(config, node):
...
+        top_dir, tail = os.path.split(node.home)
+        logger.debug('topdir: {0} --- tail: {1}'.format(top_dir, tail))
+        if 'arakoon' in top_dir and config.cluster_id in top_dir and tail == 'db':
+            root_client.dir_delete([top_dir])
```
khenderick, line 323: Why not just remove `node.log_dir` and `os.path.split(node.home)[0]` here. Not sure what the checks are about.

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -183,15 +206,23 @@ def create_cluster(cluster_name, ip, exclude_ports, base_dir, plugins=None):
         ports = System.get_free_ports(port_range, exclude_ports, 2, client)
         node_name = System.get_my_machine_id(client)
 
+        home_dir = ArakoonInstaller.ARAKOON_HOME_DIR.format(base_dir, cluster_name)
+        log_dir = ArakoonInstaller.ARAKOON_LOG_DIR.format(cluster_name)
+        tlog_dir = ArakoonInstaller.ARAKOON_TLOG_DIR.format(base_dir, cluster_name)
+
```
khenderick, line 213: It might be an idea to just create an `ArakoonInstaller.ARAKOON_BASE_DIR` containing `{0}/arakoon/{1}` so we can just rename that one to sth like `{0}/arakoon/{1}-<timestamp>` and move the contents of the (formatted) `ArakoonInstaller.ARAKOON_LOG_DIR` into that dir. That will save you calling the `archive_existing_arakoon_data` multiple times. You'll also be able to use that `ArakoonInstaller.ARAKOON_BASE_DIR` to remove the data on destroy.

PR #305, opened 2015-12-21
OVS-3812 VDiskController set_config_params

eduardBM:
- fix for single node use case

### DIFF
--- a/ovs/lib/vdisk.py
+++ b/ovs/lib/vdisk.py
@@ -611,11 +611,13 @@ class VDiskController(object):
         """
         required_params = {'dtl_mode': (str, StorageDriverClient.VDISK_DTL_MODE_MAP.keys()),
                            'sco_size': (int, StorageDriverClient.TLOG_MULTIPLIER_MAP.keys()),
-                           'dtl_target': (str, Toolbox.regex_ip),
                            'dedupe_mode': (str, StorageDriverClient.VDISK_DEDUPE_MAP.keys()),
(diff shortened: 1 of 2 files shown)


ovs/lib/vdisk.py
```diff
@@ -611,12 +611,14 @@ def set_config_params(vdisk_guid, new_config_params):
         """
         required_params = {'dtl_mode': (str, StorageDriverClient.VDISK_DTL_MODE_MAP.keys()),
                            'sco_size': (int, StorageDriverClient.TLOG_MULTIPLIER_MAP.keys()),
-                           'dtl_target': (str, Toolbox.regex_ip),
                            'dedupe_mode': (str, StorageDriverClient.VDISK_DEDUPE_MAP.keys()),
                            'write_buffer': (int, {'min': 128, 'max': 10 * 1024}),
                            'cache_strategy': (str, StorageDriverClient.VDISK_CACHE_MAP.keys()),
                            'readcache_limit': (int, {'min': 1, 'max': 10 * 1024}, False)}
 
+        if new_config_params.get('dtl_target') is not None:
```
kvanhijf, line 619: Pull request has already been closed, but as a remark for this change for future reference

You should have made this change instead:
Not removing:  'dtl_target': (str, Toolbox.regex_ip),
But adding this:  'dtl_target': (str, Toolbox.regex_ip, False),

The False at the end of the line will check if key is present. If present, it checks if it matches the regex, else it skips it

PR #329, opened 2016-01-06
OVS-3904 - Wait for arakoon collapse + Updated pyrakoon

khenderick:


ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -83,6 +83,7 @@ def __init__(self, cluster_id, plugins=None):
         self.cluster_id = cluster_id
         self._dir = ArakoonClusterConfig.ARAKOON_CONFIG_DIR.format(self.cluster_id)
         self.filename = ArakoonClusterConfig.ARAKOON_CONFIG_FILE.format(self.cluster_id)
+        self._extra_globals = {'__tainted_tlog_entries_per_file': 5000}
```
domsj, line 86: Will this only be used for new arakoon clusters and never for already existing ones?

domsj:
isn't the commit hash on which the pyrakoon import is based somewhere in the repo?

ovs/extensions/db/arakoon/ArakoonInstaller.py
khenderick, line 86: Correct, is this something that is required?

khenderick:
No, but that's a good idea to do next time I update the client.

ovs/extensions/db/arakoon/ArakoonInstaller.py
domsj, line 86: Yes, under no circumstances do you want to change this setting for an already existing cluster.
Also all nodes in the cluster should be configured in the same way regarding this property.
(Which I think from looking at your code should be the case)

khenderick, line 86: Hmm, on second thought, it will be added to existing ones as well. I'll fix this, as the idea was actually to not update this on existing clusters.

khenderick, line 86: Resolved in e325c16

PR #342, opened 2016-01-12
OVS-3960 - Move the EtcdConfiguration to the db.etcd extensions

khenderick:


khenderick:
Whoops, too fast in clicking the buttons

PR #343, opened 2016-01-12
added md support

jeroenmaelbrancke:


ovs/lib/disk.py
```diff
@@ -71,7 +71,7 @@ def sync_with_reality(storagerouter_guid=None):
                 context = remote.Context()
                 devices = [device for device in context.list_devices(subsystem='block')
                            if ('ID_TYPE' in device and device['ID_TYPE'] == 'disk') or
-                              ('DEVNAME' in device and ('loop' in device['DEVNAME'] or 'nvme' in device['DEVNAME']))]
+                              ('DEVNAME' in device and ('loop' in device['DEVNAME'] or 'nvme' in device['DEVNAME'] or 'md' in device['DEVNAME']))]
```
kvanhijf, line 74: I'd change this to 'startswith' instead of 'md' in ...
Same for loop and nvme as far as i'm concerned (slightly safer)

ovs/lib/disk.py
```diff
@@ -98,9 +100,16 @@ def sync_with_reality(storagerouter_guid=None):
                             partition_name = device_name
                             partition_id = match.groups()[1]
                             device_name = match.groups()[0]
-                    sectors = int(client.run('cat /sys/block/{0}/size'.format(device_name)))
-                    sector_size = int(client.run('cat /sys/block/{0}/queue/hw_sector_size'.format(device_name)))
-                    rotational = int(client.run('cat /sys/block/{0}/queue/rotational'.format(device_name)))
+                    if device_name.startswith('md') and 'p' in device_name:
+                        md_device = device_name[:device_name.index('p')]
```
kvanhijf: If the 'is_partition is True' on line 82 is entered, the devicename will already become device_name = device_name[:device_name.index('p')]
Will it not screw up the check on line 104?

PR #344, opened 2016-01-12
add unattended heartbeat possibility

kinvaris:


ovs/lib/setup.py
```diff
@@ -124,7 +124,7 @@ def setup_node(ip=None, force_type=None):
                 known_passwords = json.loads(config.get('setup', 'passwords'))
             if config.has_option('setup', 'external_etcd'):
                 external_etcd = json.loads(config.get('setup', 'external_etcd'))
-            enable_heartbeats = False
+            enable_heartbeats = config.getboolean('setup', 'enable_heartbeats')
```
khenderick: Can you make it optional and set the default to `False`?

wimpers: I would prefer to have the default set to true. See discussion of the BAM in OVS-3948

khenderick: Ok. but please make it optional, as we have to update all our QA environments otherwise.

ovs/lib/setup.py
```diff
@@ -124,7 +124,10 @@ def setup_node(ip=None, force_type=None):
                 known_passwords = json.loads(config.get('setup', 'passwords'))
             if config.has_option('setup', 'external_etcd'):
                 external_etcd = json.loads(config.get('setup', 'external_etcd'))
-            enable_heartbeats = False
+            try:
```
khenderick: Try again with a `config.has_option('setup', 'enable_heartbeats')` ;).

kvanhijf:
And try again with default True apparently (in the 'else' stack)   :)

kinvaris:
It should be False default right, because you would need to rewrite the QA tests?

kinvaris:
I've set the default to True, this means that you will have to inform QA that they would have to rewrite their tests if you do not want to receive heartbeats in support portal

PR #363, opened 2016-01-19
OVS-3820 - Logs to redis

khenderick:
Work in progress


khenderick:
We're not taking this road, closing

PR #368, opened 2016-01-21
Ovs 3992 fleet extension

eduardBM:


ovs/extensions/services/service.py
```diff
@@ -15,14 +15,20 @@
...
+from subprocess import check_output, CalledProcessError
 from ovs.extensions.services.upstart import Upstart
 from ovs.extensions.services.systemd import Systemd
 from ovs.log.logHandler import LogHandler
 
 logger = LogHandler.get('extensions', name='servicemanager')
-
+try:
+    from ovs.extensions.services.fleetctl import FleetCtl
+    HAS_FLEET_CLIENT = True
```
kvanhijf: No uppercase please for variables that can change
Otherwise all booleans should be uppercase

ovs/extensions/services/service.py
```diff
@@ -51,7 +57,21 @@ def __getattr__(cls, item):
                         else:
                             raise RuntimeError('The ServiceManager is unrecognizable')
                     elif 'systemd' in init_info:
-                        ServiceManager.ImplementationClass = Systemd
+                        if HAS_FLEET_CLIENT is True:
+                            try:
+                                if ServiceManager.has_fleet():
+                                    if ServiceManager._is_fleet_running_and_usable():
```
kvanhijf: if ServiceManager._is_fleet_running_and_usable() is False:
    ServiceManager.setup_fleet()

ovs/extensions/services/service.py
```diff
@@ -51,7 +57,21 @@ def __getattr__(cls, item):
                         else:
                             raise RuntimeError('The ServiceManager is unrecognizable')
                     elif 'systemd' in init_info:
-                        ServiceManager.ImplementationClass = Systemd
```
kvanhijf, line 54: Leave this line as default for systemd
then check if fleet is installed and correctly setup --> overwrite default with FleetCtl
Then you can remove all the ImplementationClass changes in all the else stacks and except stacks

ovs/extensions/services/service.py
```diff
@@ -15,14 +15,20 @@
...
-from subprocess import check_output
+from subprocess import check_output, CalledProcessError
 from ovs.extensions.services.upstart import Upstart
 from ovs.extensions.services.systemd import Systemd
 from ovs.log.logHandler import LogHandler
 
 logger = LogHandler.get('extensions', name='servicemanager')
-
+try:
+    from ovs.extensions.services.fleetctl import FleetCtl
```
kvanhijf: This try except is only interesting for systemd ==> move it in the else stack of systemd, no need to try to import on ubuntu

ovs/extensions/services/service.py
```diff
@@ -60,3 +80,54 @@ def __getattr__(cls, item):
...
+    @staticmethod
+    def reload():
+        ServiceManager.ImplementationClass = None
+
+    @staticmethod
+    def has_fleet_client():
+        return HAS_FLEET_CLIENT
+
+    @staticmethod
+    def setup_fleet():
```
kvanhijf, line 83: No use for return True or False, because you never use the returned value
Just return in case some check is not passed

ovs/extensions/services/service.py
```diff
@@ -60,3 +80,54 @@ def __getattr__(cls, item):
...
+        if HAS_FLEET_CLIENT is False:
+            logger.info('Cannot use fleet because the client is not installed')
+            return False
+        if ServiceManager.has_fleet():
+            if ServiceManager._is_fleet_running_and_usable():
+                logger.info('Fleet service is running')
+                ServiceManager.reload()
+                return True
+            else:
```
kvanhijf, line 92: No need for else stack after a return

ovs/extensions/services/service.py
```diff
@@ -60,3 +80,54 @@ def __getattr__(cls, item):
...
+            has_fleetd_bin = 'fleetd' in check_output('which fleetd', shell=True)
+            return has_fleetctl_bin and has_fleetd_bin
+        except CalledProcessError as cpe:
+            logger.warning('Could not determine if fleet can be used. {0}'.format(cpe))
+            return False
+
+    @staticmethod
+    def _is_fleet_running_and_usable():
+        try:
+            is_fleetd_running = 'active' in check_output('systemctl is-active fleet || true', shell=True).strip()
```
kvanhijf: I dont see the added value to first check if fleet is active
If you know the binaries are both present, you can try to execute the list-machines command right away --> if this succeeds return True else False, this can be a 1-liner

ovs/extensions/services/fleetctl.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+class FleetCtl(object):
+    """
+    Contains all logic related to managing services through fleet
+    """
+
+    @staticmethod
+    def prepare_template(base_name, target_name, client):
+        return Systemd.prepare_template(base_name, target_name, client)
```
kvanhijf, line 39: Unused method??

ovs/extensions/services/fleetctl.py
```diff
@@ -0,0 +1,237 @@
...
+        template_file = client.file_read(template_service.format(name))
+
+        for key, value in params.iteritems():
+            template_file = template_file.replace('<{0}>'.format(key), value)
+        if '<SERVICE_NAME>' in template_file:
+            service_name = name if target_name is None else target_name
+            template_file = template_file.replace('<SERVICE_NAME>', service_name.lstrip('ovs-'))
+
+        dependencies = ''
+        if additional_dependencies:
```
kvanhijf: if check obsolete
You should overwrite the additional dependencies with an empty list at start of function similar to what you do with params
Then you can always iterate the list, without having to check first if anything in it

ovs/extensions/services/fleetctl.py
```diff
@@ -0,0 +1,237 @@
...
+            for service in additional_dependencies:
+                dependencies += '{0}.service '.format(service)
+        template_file = template_file.replace('<ADDITIONAL_DEPENDENCIES>', dependencies)
+
+
+        template_file += "\n[X-Fleet]\nMachineID={0}".format(FleetCtl._get_id_from_ip(client_ip))
```
kvanhijf, line 79: 2 empty lines
Please respect code guidelines

ovs/extensions/services/fleetctl.py
```diff
@@ -0,0 +1,237 @@
...
+                if unit.as_dict()['currentState'] != 'launched':
+                    FLEET_CLIENT.set_unit_desired_state(unit, 'launched')
+                    time.sleep(1)
+                else:
+                    break
+            unit = FleetCtl._get_unit(fleet_name)
+            if unit.as_dict()['currentState'] != 'launched':
+                logger.warning('Failed to start unit {0}'.format(unit.as_dict()))
+            logger.debug('Fleet start unit {0} > {1}'.format(fleet_name, unit.as_dict()['currentState']))
+            return unit.as_dict()['currentState']
```
kvanhijf, line 140: Returnvalues are never used

ovs/extensions/services/fleetctl.py
```diff
@@ -0,0 +1,237 @@
...
+                if unit.as_dict()['currentState'] != 'loaded':
+                    FLEET_CLIENT.set_unit_desired_state(unit, 'loaded')
+                    time.sleep(1)
+                else:
+                    break
+            unit = FleetCtl._get_unit(fleet_name)
+            if unit['currentState'] != 'loaded':
+                logger.warning('Failed to stop unit {0}'.format(unit.as_dict()))
+            logger.debug('Fleet stop unit {0} {1}'.format(fleet_name, unit.as_dict()['currentState']))
+            return unit.as_dict()['currentState']
```
kvanhijf, line 159: Returnvalues are never used

ovs/extensions/services/fleetctl.py
```diff
@@ -0,0 +1,237 @@
...
+        return 'Service not found'
+
+    @staticmethod
+    def restart_service(name, client):
+        FleetCtl.stop_service(name, client)
+        FleetCtl.start_service(name, client)
+        return FleetCtl.get_service_status(name, client)
```
kvanhijf, line 166: Returnvalues never used

ovs/extensions/services/fleetctl.py
eduardBM, line 39: Also defined in systemd and upstart - so defined here as well. Delegated to systemd since this is not a fleet specific action.

ovs/extensions/services/fleetctl.py
eduardBM: Copied over from systemd, but makes sense to overwrite None.

ovs/extensions/services/fleetctl.py
eduardBM, line 79: Ok.

ovs/extensions/services/fleetctl.py
eduardBM, line 140: Both systemd and upstart extensions return value in this method.

ovs/extensions/services/fleetctl.py
eduardBM, line 159: Both systemd and upstart extensions return value in this method.

ovs/extensions/services/fleetctl.py
eduardBM, line 166: Both systemd and upstart extensions return value in this method.

ovs/extensions/services/service.py
eduardBM: It is interesting on newer Ubuntu which has systemd as well, but indeed, could be moved in the "else" part of service detection.

ovs/extensions/services/service.py
eduardBM: Indeed, not a constant

ovs/extensions/services/service.py
eduardBM, line 54: Ok.

ovs/extensions/services/service.py
eduardBM: Ok.

ovs/extensions/services/service.py
eduardBM, line 83: Ok.

ovs/extensions/services/service.py
eduardBM, line 92: The else is when the fleet service is detected but it's not yet running and usable, so we try to start it.

ovs/extensions/services/service.py
eduardBM: Indeed, once we can connect to it we know for sure it's active.

PR #374, opened 2016-01-26
OVS-4051 - Fix removing incomplete deployed arakoon

khenderick:


ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -142,24 +142,23 @@ def write_config(self):
         """
         Writes the configuration down to in the format expected by Arakoon
         """
-        (temp_handle, temp_filename) = tempfile.mkstemp()
```
kvanhijf, line 145: lets remove the tempfile import shall we?

khenderick, line 145: Removed, thanks

PR #383, opened 2016-01-28
OVS-3951 Don't use fdisk in sync_with_reality

eduardBM:
- issue with GPT disks

### DIFF
--- a/ovs/extensions/generic/disk.py
+++ b/ovs/extensions/generic/disk.py
@@ -81,6 +81,32 @@ class DiskTools(object):
         check_output(command_2, shell=True)
 
     @staticmethod
+    def get_partitions_info(device):
+        """
+        Get partitions info from a device
+        :param device: device name: /dev/sda
+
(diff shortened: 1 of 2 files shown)


ovs/extensions/generic/disk.py
```diff
@@ -81,6 +81,36 @@ def create_partition(disk_path, disk_size, partition_start, partition_size):
...
+        """
+        result = {}
+        command = "parted -m {0} unit B print -s".format(device)
+        logger.debug('Checking partitions with command: {0}'.format(command))
+        try:
+            output = check_output(command, shell=True).splitlines()
+        except CalledProcessError as cpe:
+            raise RuntimeError('{0} {1}'.format(cpe, cpe.output))
+        for line in output[:]:
+            output.pop(0)
```
kvanhijf, line 100: Why is this pop necessary?

eduardBM, line 100: Remove "warnings or stderr info" before "BYT;" line.
Not parseable, so removing it.

PR #395, opened 2016-02-03
OVS-3889 - DTL checkup now in deduped mode

kvanhijf:


ovs/dal/dataobjectlist.py
```diff
@@ -273,3 +281,13 @@ def pop(self, index):
...
+
+    def shuffle(self):
+        """
+        Randomly shuffle the items in the data-object list
+        :return: Shuffled data-object list
+        """
+        self.load()
+        objects = [self._objects[guid] for guid in self._guids]
+        random.shuffle(objects)
+        self._guids = [obj.guid for obj in objects]
```
khenderick: Can't you just shuffle the guids? The iterators use the order of the guids anyway and you don't have to load all objects.

ovs/dal/tests/test_basic.py
```diff
@@ -1415,6 +1415,32 @@ def raise_error():
...
+            disk.name = 'disk{0}'.format(i)
+            disk.save()
+
+        data = DataList({'object': TestDisk,
+                         'data': DataList.select.GUIDS,
+                         'query': {'type': DataList.where_operator.AND,
+                                   'items': []}}).data
+        datalist = DataObjectList(data, TestDisk)
+        starting_order = []
+        for index, disk in enumerate(datalist):
```
khenderick: `for disk in datalist` should work.

PR #397, opened 2016-02-04
OVS-4129 - Added DAL benchmark + Slower support data fetching

khenderick:


ovs/dal/dataobject.py
```diff
@@ -946,6 +946,27 @@ def __ne__(self, other):
             return True
         return not self.__eq__(other)
 
+    def _benchmark(self, iterations=100):
+        import time
+        begin = time.time()
+        stats = {}
+        totals = []
+        for i in range(iterations):
```
kvanhijf: for _ in range(iterations)

PR #409, opened 2016-02-16
Ovs 2138 rollback setup

eduardBM:


ovs/lib/setup.py
```diff
@@ -905,19 +974,30 @@ def _setup_first_node(target_client, unique_id, cluster_name, node_name, hypervi
             logger.info('Setting up fleet')
             ServiceManager.setup_fleet()
 
-        print 'Setting up Arakoon'
-        logger.info('Setting up Arakoon')
-        result = ArakoonInstaller.create_cluster('ovsdb', cluster_ip, EtcdConfiguration.get('/ovs/framework/paths|ovsdb'), locked=False)
-        arakoon_ports = [result['client_port'], result['messaging_port']]
+        try:
+            arakooncompleted = EtcdConfiguration.get('/ovs/framework/hosts/{0}/arakooncompleted'.format(unique_id)) is True
```
khenderick: `something = whatever is True` is the same as `something = whatever`, right?

ovs/lib/setup.py
```diff
@@ -905,19 +974,30 @@ def _setup_first_node(target_client, unique_id, cluster_name, node_name, hypervi
...
+        except EtcdKeyNotFound:
+            arakooncompleted = None
+        if arakooncompleted is None:
+            print 'Setting up Arakoon'
```
khenderick: If the Etcd configuration value above exists, arakoon won't be configured, even if the Etcd configuration value was `False` or even `None`. As in that case `arakooncompleted` will be `False` and this if-statement is false. If this is intended, you might rewrite this try/except and if by just using `if EtcdConfiguration.exists(<key>):`

ovs/lib/setup.py
```diff
@@ -1051,6 +1219,48 @@ def _setup_extra_node(cluster_ip, master_ip, cluster_name, unique_id, ip_client_
         logger.info('Extra node complete')
 
     @staticmethod
+    def _rollback_setup_extra_node(target_client):
```
khenderick: Can this method be merged (or at least partially) with rollback_first_node? I assume they will contain duplicate code.

ovs/lib/setup.py
```diff
@@ -1186,13 +1397,16 @@ def _demote_node(cluster_ip, master_ip, cluster_name, ip_client_map, unique_id,
...
-        if len(master_nodes) == 0:
-            raise RuntimeError('There should be at least one other master node')
+        try:
+            config = ArakoonClusterConfig('ovsdb')
+            config.load_config()
+            master_nodes = [node.ip for node in config.nodes]
+            if cluster_ip in master_nodes:
+                master_nodes.remove(cluster_ip)
+            if len(master_nodes) == 0:
+                raise RuntimeError('There should be at least one other master node')
```
khenderick: This exception is absorbed by the except below. I don't think that's intended.

ovs/lib/setup.py
```diff
@@ -1201,35 +1415,47 @@ def _demote_node(cluster_ip, master_ip, cluster_name, ip_client_map, unique_id,
         print 'Leaving arakoon ovsdb cluster'
         logger.info('Leaving arakoon ovsdb cluster')
         offline_node_ips = [node.ip for node in offline_nodes]
-        ArakoonInstaller.shrink_cluster(cluster_ip, 'ovsdb', offline_node_ips)
+        try:
+            ArakoonInstaller.shrink_cluster(cluster_ip, 'ovsdb', offline_node_ips)
+        except Exception as ex:
+            SetupController._print_log_error('leave arakoon ovsdb cluster', ex)
```
khenderick: Any idea why that might fail? It might be a better idea to make `shrink_cluster` smarter so it checks reality and just assures that once the method completes, the state is as expected (so that if you try to shrink a cluster twice, the second time it won't do anything)

ovs/lib/setup.py
```diff
@@ -1201,35 +1415,47 @@ def _demote_node(cluster_ip, master_ip, cluster_name, ip_client_map, unique_id,
...
         print 'Restarting master node services'
         logger.info('Restarting master node services')
         remaining_nodes = ip_client_map.keys()[:]
         if cluster_ip in remaining_nodes:
             remaining_nodes.remove(cluster_ip)
-        ArakoonInstaller.restart_cluster_remove('ovsdb', remaining_nodes)
+        try:
+            ArakoonInstaller.restart_cluster_remove('ovsdb', remaining_nodes)
+        except Exception as ex:
+            SetupController._print_log_error('restart arakoon cluster', ex)
```
khenderick: Any reason why that would fail? It just restarts all nodes which should always work, right?

ovs/extensions/db/etcd/installer.py
```diff
@@ -197,6 +207,9 @@ def use_external(external, slave_ip, cluster_name):
     def _setup_proxy(initial_cluster, slave_client, cluster_name):
         base_name = 'ovs-etcd-proxy'
         target_name = 'ovs-etcd-{0}'.format(cluster_name)
+        if ServiceManager.has_service(target_name, slave_client) and ServiceManager.get_service_status(target_name, slave_client) is True:
```
khenderick: If calling `shrink_cluster`, I think that this will always be `True`, so I'm not sure whether it's still possible to shrink an Etcd cluster now.

ovs/lib/setup.py
```diff
@@ -1436,10 +1659,27 @@ def _configure_logstash(client):
...
+        if ServiceManager.has_service('logstash', client) is True:
+            ServiceManager.remove_service('logstash', client)
+
+
+    @staticmethod
     def _configure_avahi(client, cluster_name, node_name, node_type):
         print '\n+++ Announcing service +++\n'
         logger.info('Announcing service')
 
+        if client.file_exists(SetupController.avahi_filename):
```
khenderick: This method is also used for reconfiguring Avahi. So if it was already configured as `master`, and it's called to reconfigure it as `extra`, it will skip this, right?

ovs/lib/setup.py
```diff
@@ -1458,6 +1698,12 @@ def _configure_avahi(client, cluster_name, node_name, node_type):
         Toolbox.change_service_state(client, 'avahi-daemon', 'restart', logger)
 
     @staticmethod
+    def _unconfigure_avahi(client):
```
khenderick: Is this one used?

PR #423, opened 2016-02-23
Add healthcheck easy execution command

kinvaris:
Also OVS healthcheck (currently first release) should be shipped with the 'openvstorage' package.


kvanhijf:
Add summary on top of the file similar to any other possible action

kinvaris:
This ticket can be closed because it is integrated with: #473

PR #424, opened 2016-02-24
Do not raise an error if the file already exists during create_volume

jamie-liu:


eduardBM:
I would strongly advise against this.
There are several scenarios that would break due to this:
- you must not be able to create two volumes with same name (since the volumedriver vpool is a filesystem, you cannot have two files with same name)
- allowing to call "truncate -s XG" again could have several outcomes:
  -- if the size is the same, the command will return without creating a new volume, thus telling Cinder the volume is created, although the volume was there (either created for another vm and running, or from a previous failed attempts)
  -- if the size is greater than the existing file will cause the volumedriver to increase the volume in size (which would impact the FS on the volume, and the OS running on it), and will tell Cinder the volume is created (see above)
  -- if the size is smaller, then the call will fail.

So, the correct solution would be to keep this check, and make sure a failed attempt always removes the .raw file (which it should, by calling "VDiskController.delete_volume" - https://github.com/openvstorage/framework-cinder-plugin/blob/master/cinder-volume-driver/kilo/openvstorage.py#L132)

domsj:
Would it help if an optional parameter `fail_if_exists = True` were added to this method?
Then jamie could only update the specific caller he has in mind.

jamie-liu:
Hi Eddie,

A direct return here when File already exists will not allow to truncate  a new file with same name anymore. 
In our case, we did see the _find_ovs_model_disk_by_location raise an error in the first try which lead to the error below when Cinder retry to create the volume:
I think the better way is to root cause the _find_ovs_model_disk_by_location  error instead of change this code.

2016-02-22 09:44:31.334 25576 TRACE oslo_messaging.rpc.dispatcher   File "/usr/lib/python2.7/site-packages/cinder/volume/drivers/openvstorage.py", li
ne 125, in create_volume
2016-02-22 09:44:31.334 25576 TRACE oslo_messaging.rpc.dispatcher     size = size)
2016-02-22 09:44:31.334 25576 TRACE oslo_messaging.rpc.dispatcher   File "/usr/lib/python2.7/site-packages/celery/local.py", line 188, in call
2016-02-22 09:44:31.334 25576 TRACE oslo_messaging.rpc.dispatcher     return self._get_current_object()(_a, *_kw)
2016-02-22 09:44:31.334 25576 TRACE oslo_messaging.rpc.dispatcher   File "/usr/lib/python2.7/site-packages/celery/app/task.py", line 420, in call
2016-02-22 09:44:31.334 25576 TRACE oslo_messaging.rpc.dispatcher     return self.run(_args, *_kwargs)
2016-02-22 09:44:31.334 25576 TRACE oslo_messaging.rpc.dispatcher   File "/opt/OpenvStorage/ovs/lib/vdisk.py", line 541, in create_volume
2016-02-22 09:44:31.334 25576 TRACE oslo_messaging.rpc.dispatcher     raise RuntimeError('File already exists at %s' % location)
2016-02-22 09:44:31.334 25576 TRACE oslo_messaging.rpc.dispatcher RuntimeError: File already exists at /mnt/bjdxtovs/volume-e837ca61-ff50-4fc1-a995-0d8ca57bf205.raw

kvanhijf:
Proposed solution is not a fix, but workaround instead
Instead we should indeed (as Jamie pointed out) investigate why something went wrong creating the volume.
1st things that comes to mind for me are the length of the volumename perhaps or the dashes in its name?

eduardBM:
@jamie-liu 
Indeed, the correct solution is to find the cause of the initial fail and why the cleanup failed in the first place.
More info on this can be found in ovs-workers logs.

PR #432, opened 2016-02-26
Ovs 4176 service check 270

eduardBM:


ovs/extensions/services/systemd.py
```diff
@@ -92,13 +92,19 @@ def add_service(name, client, params=None, target_name=None, additional_dependen
             raise Exception('Add {0}.service failed, {1}'.format(name, output))
 
     @staticmethod
-    def get_service_status(name, client):
+    def get_service_status(name, client, return_output=False):
```
kvanhijf, line 95: Function returns tuple or boolean, while this is possible in Python, we should not do this.
Imo, we should always return status + output and adapt all get_service_status calls where a boolean is expected as only returnvalue

PS: autotests also rely on this call, plz fix those too

PR #433, opened 2016-02-26
Ovs 3909 gui option create vdisk

pploegaert:


webapps/api/backend/views/vdisks.py
```diff
@@ -159,6 +159,28 @@ def set_as_template(self, vdisk):
...
     @required_roles(['read', 'write'])
     @return_task()
+    @load(StorageRouter)
+    def create(self, devicename, size, vpool_guid, storagerouter_guid):
+        """
+        Create a new vdisk
+        :param devicename: Name of the new vdisk
+        :param size: size of  virtual disk
+        :param vpool_guid: Guid of vPool to create new vdisk on
+        :param storagerouter_guid: Guid of the storagerouter_guid to assign disk to
```
kvanhijf: Guid of the storagerouter

webapps/frontend/app/viewmodels/wizards/addvdisk/data.js
```diff
@@ -0,0 +1,48 @@
...
+            storageRouter:  ko.observable(),
+            storageRouters: ko.observableArray([]),
+            vPool:          ko.observable(),
+            vPools:         ko.observableArray([])
+        };
+
+        // Computed
+        wizardData.size = ko.computed( function () {
+            var size = wizardData.size_entry();
+            if (wizardData.size_unit() === 'gib') {
```
kvanhijf: gib implies multiplication by 1024
gb implies multiplication by 1000

webapps/frontend/app/viewmodels/wizards/addvdisk/gather.js
```diff
@@ -0,0 +1,131 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+/*global define */
+define([
+    'jquery', 'knockout',
+    'ovs/shared', 'ovs/api', 'ovs/generic', './data', '../../containers/vpool', '../../containers/storagerouter'
+], function($, ko, shared, api, generic, data, VPool, StorageRouter) {
```
kvanhijf: shared is unused

webapps/frontend/app/viewmodels/wizards/addvdisk/gather.js
```diff
@@ -0,0 +1,131 @@
...
+                .done(function(data) {
+                    var guids = [], srdata = {};
+                    $.each(data.data, function(index, item) {
+                        guids.push(item.guid);
+                        srdata[item.guid] = item;
+                    });
+                    generic.crossFiller(
+                        guids, self.data.storageRouters,
+                        function(guid) {
+                            if (self.data.vPool() === undefined || !self.data.vPool().storageRouterGuids().contains(guid)) {
```
kvanhijf: At this point self.data.vPool() will always be undefined
Which will result in all storagerouters visible, which will allow the customer to select a SR not linked to the current vPool

webapps/frontend/app/viewmodels/wizards/addvdisk/gather.js
```diff
@@ -0,0 +1,131 @@
...
+            }
+
+        }
+        self.finish = function() {
+            return $.Deferred(function(deferred) {
+                generic.alertInfo(
+                    $.t('ovs:wizards.add_vdisk.gather.started'),
+                    $.t('ovs:wizards.add_vdisk.gather.inprogress', { what: self.data.name() })
+                );
+                deferred.resolve();
```
kvanhijf: If you dont want to wait for the disk to be created, its still best to resolve afer launching the task

webapps/frontend/app/viewmodels/wizards/addvdisk/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+        // Setup
+        self.title(generic.tryGet(options, 'title', $.t('ovs:wizards.add_vdisk.title')));
+        self.modal(generic.tryGet(options, 'modal', false));
+
+        self.steps([new Gather()]);
+        self.activateStep();
+
+        // Cleaning data
+        data.name('');
```
kvanhijf, line 36: Will filling out 'size' in the wizard, leaving the wizard and re-entering the wizard, clear the 'size' field?   Don't know for sure
Same for storagerouter, this won't be reset either, which might not be an issue

webapps/frontend/app/views/wizards/addvdisk/gather.html
```diff
@@ -0,0 +1,81 @@
...
+    http://www.openvstorage.org/license
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<div>
+    <form role="form form-horizontal" class="form-horizontal">
```
kvanhijf: 'role' in a form should be removed, pycharm gives a warning about this

webapps/frontend/app/views/wizards/addvdisk/gather.html
```diff
@@ -0,0 +1,81 @@
...
+        <div class="form-group" data-bind="css: { 'has-error': $.inArray('name', canContinue().fields) !== -1 }">
```
kvanhijf: Also here, replace $.inArray with contains please, more readable

webapps/frontend/app/viewmodels/wizards/addvdisk/gather.js
```diff
@@ -0,0 +1,131 @@
...
+                            $.t('ovs:wizards.add_vdisk.gather.complete'),
+                            $.t('ovs:wizards.add_vdisk.gather.success', { what: self.data.name() })
+                        );
+                    })
+                    .fail(function(error) {
+                        error = $.parseJSON(error.responseText);
+                        generic.alertError(
+                            $.t('ovs:generic.error'),
+                            $.t('ovs:wizards.add_vdisk.gather.failed', {
+                                what: self.data.name(),
```
kvanhijf: The 'failed' entry in the ovs.json does not have a what, nor a why

kvanhijf:
As a general remark regarding this review:
From what i see, its currently possible to select a vpool and a storagerouter, which are not linked to eachother.

webapps/frontend/app/viewmodels/wizards/addvdisk/gather.js
```diff
@@ -0,0 +1,131 @@
...
+                        devicename: self.data.name(),
+                        size: self.data.size(),
+                        vpool_guid: self.data.vPool().guid(),
+                        storagerouter_guid: self.data.storageRouter().guid()
+                    }
+                })
+                    .done(function() {
+                        generic.alertSuccess(
+                            $.t('ovs:wizards.add_vdisk.gather.complete'),
+                            $.t('ovs:wizards.add_vdisk.gather.success', { what: self.data.name() })
```
kvanhijf: Success also does not have a 'what' in ovs.json, neither does inprogress

webapps/frontend/app/viewmodels/wizards/addvdisk/gather.js
```diff
@@ -0,0 +1,131 @@
...
+            })
+                .done(function(data) {
+                    var guids = [], srdata = {};
+                    $.each(data.data, function(index, item) {
+                        guids.push(item.guid);
+                        srdata[item.guid] = item;
+                    });
+                    generic.crossFiller(
+                        guids, self.data.storageRouters,
+                        function(guid) {
```
khenderick: You _must_ return an object in this function. Not returning anything will cause strange behavior (or undefined objects - which will make other code more complex to handle this). If there's a guid which you don't want included, you need to filter these out before adding them to the `crossFiller`. So basically move the below if-statement to `$.each(...)` above.

webapps/api/backend/views/vdisks.py
```diff
@@ -159,6 +159,28 @@ def set_as_template(self, vdisk):
     @log()
     @required_roles(['read', 'write'])
     @return_task()
+    @load(StorageRouter)
```
khenderick: You're loading a StorageRouter here. That doesn't make sense, since you're at the `create` method of the vDisk API. This should just be `@load()` without any argument. I would even expect it to fail.

khenderick: Correction, it won't fail, but it will ignore it, so it can better be left out.

webapps/frontend/app/viewmodels/wizards/addvdisk/data.js
```diff
@@ -0,0 +1,56 @@
...
+            var size = wizardData.size_entry();
+            if (wizardData.size_unit() === 'tib') {
+                size *= 1024;
+            }
+            return size;
+        });
+
+        wizardData.storageRoutersByVpool = ko.computed(function() {
+            var i, result = [];
+            for (i = 0; i < wizardData.storageRouters().length; i++) {
```
khenderick: Please use sth like the below snippet as we use this almost everywhere:

`$.each(wizardData.storageRouters(), function(index, storagerouter) {`

webapps/frontend/app/viewmodels/wizards/addvdisk/gather.js
```diff
@@ -0,0 +1,123 @@
...
+    "use strict";
+    return function() {
+        var self = this;
+
+        // Variables
+        self.data   = data;
+        self.loadStorageRoutersHandle = undefined;
+        self.loadvPoolsHandle = undefined;
+
+        // Observables
```
khenderick: Some small cleanup

webapps/frontend/app/viewmodels/wizards/addvdisk/gather.js
```diff
@@ -0,0 +1,123 @@
...
+
+        // Computed
+        self.canContinue = ko.computed(function() {
+            var valid = true, showErrors = false, reasons = [], fields = [];
+            if (self.data.name() === undefined || self.data.name() === '' || !self.data.name.valid()) {
+                valid = false;
+                fields.push('name');
+                reasons.push($.t('ovs:wizards.add_vdisk.gather.invalid_name'));
+            }
+            return { value: valid, showErrors: showErrors, reasons: reasons, fields: fields };
```
khenderick, line 45: Shouldn't we check here if e.g. `self.data.vPool()` isn't undefined, as this is used in the finish function as well.

webapps/frontend/app/viewmodels/wizards/addvdisk/gather.js
```diff
@@ -0,0 +1,123 @@
...
+                        );
+                        $.each(self.data.vPools(), function (index, vpool) {
+                            if (guids.contains(vpool.guid())) {
+                                vpool.fillData(vpData[vpool.guid()]);
+                            }
+                        });
+                    });
+                self.loadStorageRoutersHandle = api.get('storagerouters', {
+                    queryparams: {
+                        contents: 'storagedrivers,vpools,vpools_guids',
```
khenderick: `vpools_guids` should be removed here.

webapps/frontend/app/views/wizards/addvdisk/gather.html
```diff
@@ -0,0 +1,81 @@
...
+        <div class="form-group">
+            <label class="col-lg-3 control-label" data-i18n="ovs:wizards.add_vdisk.gather.vpool"></label>
+            <div class="col-lg-7">
+                <span data-bind="dropdown: {
+                                    key: 'guid',
+                                    keyisfunction: true,
+                                    items: data.vPools,
+                                    target: data.vPool,
+                                    text: function(item) {
+                                        if (item !== undefined) {
```
khenderick, line 48: What's in this dropdown if the vPools are still loading?

webapps/frontend/app/viewmodels/wizards/addvdisk/index.js
khenderick, line 36: It's a computed property, so it should be re-generated by the below two items.

webapps/frontend/app/views/wizards/addvdisk/gather.html
```diff
@@ -0,0 +1,81 @@
...
+                                            return item.name()
+                                        }
+                                    }
+                                }">
+                </span>
+            </div>
+        </div>
+        <div class="form-group">
```
khenderick: This div should have the css `has-error` css binding when the StorageRouter is empty. Similar for the vPool,

webapps/frontend/app/views/wizards/addvdisk/gather.html
khenderick, line 48: As discussed with @pploegaert, it correctly indicates the data is loading.

PR #435, opened 2016-02-26
OVS-4265 check if snapshot has no clones before deleting it

kinvaris:
in ovs-scheduled-tasks you need to check if a snapshot has clones before deleting it.


kinvaris:
this needs to be commited to unstable

khenderick:
Sorry, I have to reject this; this code is in the progress of a major rewrite, so it doesn't make sense to merge it in, to only introduce merge issues.

PR #440, opened 2016-02-29
OVS-4276 - Variable cluster size

kvanhijf:


kvanhijf:
Still a WIP
Initial commit was to be able to install volumedriver dev build

PR #453, opened 2016-03-02
OVS-3756-metadata-cache-size-on-vdisk

pploegaert:


ovs/lib/vdisk.py
```diff
@@ -62,6 +62,8 @@
 storagerouterclient.Logger.setupLogging(LogHandler.load_path('storagerouterclient'))
 storagerouterclient.Logger.enableLogging()
 
+METADATA_CACHE_PAGE_SIZE = 256 * 24
```
khenderick: I'd add these to the `StorageDriverClient`, as it currently already contains a lot of stuff like defaults etc.

ovs/lib/vdisk.py
```diff
@@ -780,6 +791,8 @@ def set_config_params(vdisk_guid, new_config_params):
                         block_size = vol_info.lba_size * vol_info.cluster_multiplier or 4096
                         limit = new_value * 1024 * 1024 * 1024 / block_size if new_value else None
                         vdisk.storagedriver_client.set_readcache_limit(volume_id, limit)
+                    elif key =='metadata_cache_size':
+                        vdisk.storagedriver_client.set_metadata_cache_capacity(str(vdisk.volume_id), int(new_value / METADATA_CACHE_PAGE_SIZE))
```
kvanhijf: str(vdisk.volume_id) can be replaced by volume_id (similar to the other set methods)
int(new_value / METADATA...) is obsolete cast, new_value has been checked by verify_required_params to be an int and METADATA... is an int

PR #454, opened 2016-03-02
OVS-4128 - Created logrotation test

kvanhijf:


ovs/extensions/generic/sshclient.py
```diff
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+SSHClient class
```
khenderick: It's a module

PR #456, opened 2016-03-04
OVS-3791-schedule-backend-sync

pploegaert:


webapps/api/backend/views/vdisks.py
```diff
@@ -232,3 +232,16 @@ def delete(self, vdisk):
...
+    @action()
+    @log()
+    @required_roles(['read', 'write'])
+    @return_task()
+    @load(VDisk)
+    def schedule_backend_sync(self, vdisk):
+        """
+        Schedule a backend sync on a vdisk
+        :param vdisk: vdisk to schedule a backend sync to
+        :return: TLogName associated with the data sent off to the backend
```
kvanhijf, line 245: Returns a celery asyncresult actually

webapps/api/backend/views/vdisks.py
```diff
@@ -232,3 +232,16 @@ def delete(self, vdisk):
...
+        """
+        return VDiskController.schedule_backend_sync.delay(vdisk_guid=vdisk.guid)
```
kvanhijf: Just to be sure, but the backend sync should not be triggered on the storagerouter hosting the volume?

PR #461, opened 2016-03-08
OVS-4289 Error in get_metadata

eduardBM:
- fix du command

### DIFF
--- a/ovs/lib/storagerouter.py
+++ b/ovs/lib/storagerouter.py
@@ -112,7 +112,7 @@ class StorageRouterController(object):
                     if client.dir_exists(storagedriver_partition.path):
                         try:
                             used_size, _ = client.run('du -B 1M -d 0 {0}'.format(storagedriver_partition.path)).split('\t')
-                            directory_used_size = int(used_size) * 1024
+                            directory_used_size = int(used_size)
                         except Exception as ex:
                             logger.error('Failed to get directory usage for {0}. {1}'.format(storagedriver_partition.path, ex))
                     used_space_by_roles += directory_used_size
--- a/ovs/lib/storagerouter.py
+++ b/ovs/lib/storagerouter.py
@@ -112,18 +112,10 @@ class StorageRouterController(object):
                     if client.dir_exists(storagedriver_partition.path):
                         try:
                             output = client.run('du -d 0 {0}'.format(storagedriver_partition.path))
+                            _used_size, _ = output.split('\t')
+                            _directory_used_size = int(_used_size) * 1024
                         except Exception as ex:
                             logger.error('Failed to get directory usage for {0}. {1}'.format(storagedriver_partition.path, ex))
-                        else:
-                            try:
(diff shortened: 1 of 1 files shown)


ovs/lib/storagerouter.py
```diff
@@ -109,20 +109,21 @@ def get_metadata(storagerouter_guid):
...
-                        output = client.run('du -Pl {0}'.format(storagedriver_partition.path))
-                    except Exception as ex:
-                        logger.error('Failed to get directory usage for {0}. {1}'.format(storagedriver_partition.path, ex))
-                    else:
+                    if client.dir_exists(storagedriver_partition.path):
                         try:
-                            _used_size, _ = output.split('\t')
-                        except ValueError as ve:
-                            logger.error('Unexpected output for directory {0}. "{1}" . {2}'.format(storagedriver_partition.path, output, ve))
+                            output = client.run('du -d 0 {0}'.format(storagedriver_partition.path))
```
kvanhijf: Eduard, while you are making changes to this code, is there any chance you could optimize this further?  Cause tbh this is very badly written code
3 try excepts after each other each with an extra indentation
- try except else statements are very unreadable for people not used to python

khenderick: I agree with @kvanhijf. This code should be made a lot shorter. For example, sth like this:

```
if client.dir_exists(foo):
    try:
        output = client.run('du -d 0 {0}'.format(foo))
        output_size, _ = output.split('\t')
        directory_used_size = int(output_size) * 1024
    except Exception as ex:
        logger.exception('Could not find disk usage: {0}'.format(ex))
        # do something, can the code below still be used?
```

If you want separate exception messages, you could catch specific errors (CalledProcessError, ValueError, ...) and log a separate message, and you could check the length of the `output.split(...)` before unpacking it (which can cause a ValueError).

kvanhijf: Don't know whether this should even be in a try - except
First you check if the path exists and if it does i dont see any reason why the 'du' command would even fail
If it would fail, something else and much more severe will be wrong imo

ovs/lib/storagerouter.py
```diff
@@ -108,22 +108,14 @@ def get_metadata(storagerouter_guid):
...
-                        except ValueError as ve:
-                            logger.error('Unexpected output for directory {0}. "{1}" . {2}'.format(storagedriver_partition.path, output, ve))
-                        else:
-                            try:
-                                _directory_used_size = int(_used_size) * 1024
-                            except ValueError as ve:
-                                logger.error('Could not parse value: {0}. {1}'.format(_used_size, ve))
-                    used_space_by_roles += _directory_used_size
+                            used_size, _ = client.run('du -B 1M -d 0 {0}'.format(storagedriver_partition.path)).split('\t')
+                            directory_used_size = int(used_size) * 1024
```
kvanhijf: You still do the multiplication?
Using the -B 1M should already return the value in megabytes and this is what you need right?

PR #464, opened 2016-03-09
OVS-4296 Route task to specified storagerouter

eduardBM:


webapps/api/backend/views/vdisks.py
```diff
@@ -171,9 +171,10 @@ def create(self, devicename, size, vpool_guid, storagerouter_guid):
         storagerouter = StorageRouter(storagerouter_guid)
         for storagedriver in storagerouter.storagedrivers:
             if storagedriver.vpool.guid == vpool_guid:
-                return VDiskController.create_new.delay(diskname=devicename,
-                                                        size=size,
-                                                        storagedriver_guid=storagedriver.guid)
+                return VDiskController.create_new.s(diskname=devicename,
```
kvanhijf: Declined commit as discussed on Skype chat
Task should not be routed. In the lower layer the vdiskcontroller should route the command itself

ovs/lib/vdisk.py
```diff
@@ -526,26 +527,37 @@ def create_new(diskname, size, storagedriver_guid):
...
+            output = check_output('truncate -s {0}G "{1}"'.format(size, location), shell=True).strip()
+            output = output.replace('\xe2\x80\x98', '"').replace('\xe2\x80\x99', '"')
 
-        if not os.path.exists(location):
-            raise RuntimeError('Cannot create file %s. Output: %s' % (location, output))
+            if not os.path.exists(location):
+                raise RuntimeError('Cannot create file %s. Output: %s' % (location, output))
+        else:
+            storagerouter = StorageRouter(storagerouter_guid)
+            client = SSHClient(storagerouter.ip)
```
kvanhijf: No issue, but just a tip
You can pass a storagerouter object to the SSHClient, no IP required

ovs/lib/vdisk.py
```diff
@@ -526,26 +527,37 @@ def create_new(diskname, size, storagedriver_guid):
...
 
         @param location: location, filename
         @param size: size of volume, GB
+        @param: storagerouter_guid: use ssh client to create file on remote storagerouter
         @return None
         """
         logger.info('Creating volume {0} of {1} GB'.format(location, size))
-        if os.path.exists(location):
-            raise RuntimeError('File already exists at %s' % location)
+        if storagerouter_guid is None:
```
kvanhijf: I'd merge the 2 paths to prevent code duplication
client = SSHClient(storagerouter) if storagerouter is not None else SSHClient('127.0.0.1')

Then you can execute all calls using the client and the SSHClient will use check_output itself when executed locally

PR #465, opened 2016-03-10
OVS-4358 - GeoScale

kvanhijf:
Initial commit to create branch
Removed possibility to run setup remote
WIP


kvanhijf:
Forgot to add comment to my commit, so here it is
Implementation of external etcd in setup.py

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -414,7 +437,11 @@ def _get_free_ports(client):
                             exclude_ports.append(node.messaging_port)
                 except:
                     logger.error('  Could not load port information of cluster {0}'.format(cluster_name))
-        ports = System.get_free_ports(EtcdConfiguration.get('/ovs/framework/hosts/{0}/ports|arakoon'.format(node_name)), exclude_ports, 2, client)
+        key = '/ovs/framework/hosts/{0}/ports|arakoon'.format(node_name)
+        if EtcdConfiguration.exists(key):
+            ports = System.get_free_ports(EtcdConfiguration.get(), exclude_ports, 2, client)
```
khenderick: You're missing key here.

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -187,6 +190,26 @@ def __init__(self):
         raise RuntimeError('ArakoonInstaller is a complete static helper class')
 
     @staticmethod
+    def get_etcd_config_path():
+        return ArakoonInstaller.ETCD_CONFIG_PATH
```
khenderick: What's the purpose of these methods? You could just simply get or set the property itself. And if you still need a setter and getter, it might be a good idea to actually use setters and getters.

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -719,3 +722,64 @@ def restart_cluster_remove(cluster_name, remaining_ips):
...
+        """
+
+        if ArakoonInstaller.is_internal(cluster_type):
+            key = ArakoonInstaller.metadata_key.format(cluster_name)
+            if cluster_name == '':
+                raise RuntimeError('Cluster name required when marking internal cluster')
+            if not EtcdConfiguration.exists(key):
+                raise RuntimeError('No cluster metadata found for {0}'.format(cluster_name))
+            values = json.loads(EtcdConfiguration.get(key, raw=True))
+            if values['in_use'] == True:
```
khenderick: you could use `EtcdConfiguration.get(key)` that would do the `json.loads(...)` etc for you. Even more, you can also use stuff like `EtcdConfiguration.get('/ovs/foo/bar|some.key')` which will even traverse the json object and return the value of that key. The set works exactly the same. More documentation and examples are in the EtcdConfiguration class' docstring.

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -244,19 +319,44 @@ def clean_leftover_arakoon_data(ip, directories):
...
         """
+        if cluster_type not in ServiceType.ARAKOON_CLUSTER_TYPES:
+            raise ValueError('Cluster type {0} is not supported. Please choose from {1}'.format(cluster_type, ', '.join(ServiceType.ARAKOON_CLUSTER_TYPES)))
+
         logger.debug('Creating cluster {0} on {1}'.format(cluster_name, ip))
         base_dir = base_dir.rstrip('/')
 
-        client = SSHClient(ip)
+        EtcdConfiguration.set('/ovs/framework/stores', {'persistent': 'pyrakoon',
+                                                        'volatile': 'memcache'})
```
khenderick: It doesn't make sense to overwrite the volatile configuration when you set up a new cluster.

khenderick: Will this be executed even if the customer wants to use external clusters set up by some whatever script using whatever other code?

The issue is that this piece is not related to arakoon at all. If the issue is that these keys are required for the arakooninstaller extension to work, why not just execute the etcd installer's initialize method before start using the arakooninstaller?

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -390,14 +517,68 @@ def deploy_cluster(cluster_name, node_ip):
...
+        :type in_use: bool
+
+        :return: List of ArakoonClusterMetadata objects
+        :rtype: list
+        """
+        clusters = []
+        cluster_type = cluster_type.upper()
+        if cluster_type not in ServiceType.ARAKOON_CLUSTER_TYPES:
+            raise ValueError('Unsupported arakoon cluster type provided. Please choose from {0}'.format(', '.join(ServiceType.ARAKOON_CLUSTER_TYPES)))
+        if not EtcdConfiguration.exists('/ovs/arakoon', raw=True):
```
khenderick: Maybe `dir_exists` is more appropriate here.

ovs/extensions/db/arakoon/ArakoonInstaller.py
```diff
@@ -414,7 +595,11 @@ def _get_free_ports(client):
                             exclude_ports.append(node.messaging_port)
                 except:
                     logger.error('  Could not load port information of cluster {0}'.format(cluster_name))
-        ports = System.get_free_ports(EtcdConfiguration.get('/ovs/framework/hosts/{0}/ports|arakoon'.format(node_name)), exclude_ports, 2, client)
+        key = '/ovs/framework/hosts/{0}/ports|arakoon'.format(node_name)
```
khenderick: This change might become obsolete if you call the etcdconfiguration's initialize before you start using the arakooninstaller. Then the `ARAKOON_START_PORT` isn't needed either. And you have the benefit that you might be able to adapt the initialize method to overwrite arakoon ports before, so you can setup clusters on other ports.

ovs/extensions/storage/persistentfactory.py
```diff
@@ -30,16 +30,25 @@ class PersistentFactory(object):
...
         if not hasattr(PersistentFactory, 'store') or PersistentFactory.store is None:
             if client_type is None:
                 client_type = EtcdConfiguration.get('/ovs/framework/stores|persistent')
 
+            framework_cluster_name = 'ovsdb'
+            for cluster in EtcdConfiguration.list('/ovs/arakoon'):
+                metadata_key = '/ovs/arakoon/{0}/metadata'.format(cluster)
+                if EtcdConfiguration.exists(metadata_key):
+                    metadata = EtcdConfiguration.get(metadata_key)
+                    if metadata['type'].upper() == 'FWK':
```
khenderick: Use ArakoonMetadata object and use the enumerators. I'd place this in the correct if below as well.

ovs/extensions/storageserver/storagedriver.py
```diff
@@ -176,6 +176,9 @@ def load(service):
         :param service: Service for which the MDSClient needs to be loaded
         """
 
+        if service.storagerouter is None:
+            return None
```
khenderick: It's better to raise an error here. This only should happen when we made a logical error in our code so we should see that error fairly soon. If we return None, it might go unnoticed.

ovs/lib/mdsservice.py
```diff
@@ -636,7 +668,7 @@ def get_mds_storagedriver_config_set(vpool, check_online=False):
                     client.run('pwd')
                 except UnableToConnectException:
                     continue
-            mds_service, load = MDSServiceController.get_preferred_mds(storagerouter, vpool, include_load=True)
+            mds_service, load = MDSServiceController.get_preferred_mds(storagerouter, vpool)
```
khenderick, line 675: A TypeError error might be raised here when None is returned. It might be better to either validate the return value, or always return a tuple.

ovs/lib/storagedriver.py
```diff
@@ -120,31 +129,44 @@ def on_demote(cluster_ip, master_ip, offline_node_ips=None):
...
+            if service.name == 'arakoon-voldrv' and service.is_internal is True:  # Externally managed arakoon cluster service does not have storage router
                 if service.storagerouter.ip == cluster_ip:
                     current_service = service
                 elif service.storagerouter.ip not in offline_node_ips:
                     remaining_ips.append(service.storagerouter.ip)
         if current_service is not None:
             print '* Shrink StorageDriver cluster'
-            ArakoonInstaller.shrink_cluster(cluster_ip, 'voldrv', offline_node_ips)
+            clusters = ArakoonInstaller.get_arakoon_metadata_by_cluster_type(cluster_type=ServiceType.ARAKOON_CLUSTER_TYPES.SD, in_use=True)
+            if len(clusters) != 1:
```
khenderick: Should check for `> 0`.

ovs/lib/storagedriver.py
```diff
@@ -165,71 +187,95 @@ def manual_voldrv_arakoon_checkup():
...
-        if 0 < len(current_services) < len(available_storagerouters):
+
+        if create_cluster is True and len(claimed_clusters) == 0 and len(current_services) == 0:  # Create new cluster
+            metadata = ArakoonInstaller.get_arakoon_metadata_by_cluster_type(cluster_type=ServiceType.ARAKOON_CLUSTER_TYPES.SD, in_use=False)
+            if len(metadata) == 0:  # No externally managed cluster found, we create 1 ourselves
+                if not available_storagerouters:
+                    raise RuntimeError('Could not find any Storage Router with a DB role')
+
+                storagerouter, partition = available_storagerouters.items()[0]
+                cluster_name = 'voldrv'
```
khenderick: You should make sure that if another cluster would be called 'voldrv' that this code can either generate another name and continue, or that the create_cluster raises that such cluster already exists. Now it might just silently mess up configurations.

ovs/lib/storagedriver.py
```diff
@@ -165,71 +187,95 @@ def manual_voldrv_arakoon_checkup():
...
+            else:
+                ports = []
+                storagerouter = None
+
+            if len(metadata) == 0:
+                raise ValueError('No arakoon cluster found of type {0}'.format(ServiceType.ARAKOON_CLUSTER_TYPES.SD))
+
+            cluster_name = metadata[0].cluster_id
+            logger.info('Claiming {0} managed arakoon cluster: {1}'.format('externally' if storagerouter is None else 'internally', cluster_name))
+            metadata[0].claim()
```
khenderick: We should have locking here. This cluster might already be claimed by another codepath.

ovs/lib/storagerouter.py
```diff
@@ -371,7 +378,7 @@ def add_vpool(parameters):
                 backend_name = backend_info['name']
                 if preset_name not in [preset['name'] for preset in backend_info['presets']]:
                     raise RuntimeError('Given preset {0} is not available in backend {1}'.format(preset_name, backend_guid))
-                task_id = ovs_client.get('/alba/backends/{0}/get_config_metadata'.format(backend_guid))
+                task_id = ovs_client.get('/alba/backends/{0}/get_arakoon_config'.format(backend_guid))
```
khenderick: I assume you have obsoleted the old call with versions?

ovs/lib/tests/test_mdsservice.py
```diff
@@ -227,7 +227,7 @@ def _create_vdisks_for_mds_service(self, amount, start_id, mds_service=None, vpo
         """
         vdisks = {}
         storagedriver_id = None
-        if mds_service is not None:
+        if mds_service is not None and mds_service.service.storagerouter_guid is not None:
```
khenderick: It should use the dynamic property, and is this even required? It looks like a silent ignore here, as the storagerouter is mandatory for an MDS service.

ovs/extensions/generic/watcher.py
```diff
@@ -109,11 +107,10 @@ def services_running(target):
             tries = 0
             while tries < max_tries:
                 try:
+                    from ovs.extensions.db.etcd.configuration import EtcdConfiguration
                     from ovs.extensions.storage.persistent.pyrakoonstore import PyrakoonStore
-                    clusters = ArakoonInstaller.get_arakoon_metadata_by_cluster_type(cluster_type=ServiceType.ARAKOON_CLUSTER_TYPES.SD, in_use=True)
-                    if len(clusters) != 1:
-                        raise ValueError('Expected exactly 1 "{0}" arakoon cluster'.format(ServiceType.ARAKOON_CLUSTER_TYPES.SD))
-                    client = PyrakoonStore(cluster=str(clusters[0].cluster_id))
+                    cluster_name = str(EtcdConfiguration.get('/ovs/framework/arakoon_clusters|voldrv'))
```
khenderick, line 112: Double casting to string

khenderick:
We decided to resolve a few more small entries in this ticket, removing the approved label.

ovs/lib/setup.py
```diff
@@ -258,33 +257,160 @@ def setup_node(node_type=None):
                 if avahi_installed is True and cluster_name is None:
                     raise RuntimeError('The name of the cluster should be known by now.')
 
-                node_password = master_password
-                for node_info in SetupController.nodes.itervalues():
+                ip_hostname_map = {cluster_ip: list({node_name, fqdn_name})}
+                for node_host_name, node_info in SetupController.nodes.iteritems():
```
khenderick, line 261: You don't need to loop here

kvanhijf, line 261: Actually i do need to loop, because i don't know the hostname of the master node
And i need the hostname to put in the ip_hostname_map
--> will still add a break in if statement

PR #466, opened 2016-03-10
Update README.md

saelbrec:
After rename to framework


wimpers:
Should we not point the release notes to the home wiki?

PR #467, opened 2016-03-11
OVS-4095 Check for duplicate names when creating from template

eduardBM:


webapps/frontend/app/viewmodels/wizards/createfromtemplatewizard/gather.js
```diff
@@ -66,6 +67,26 @@ define([
...
+                valid = false;
+                fields.push('name');
+                reasons.push($.t('ovs:wizards.createft.gather.duplicatename', {
+                    name: self.data.name()
+                }));
+            }
+            if (self.data.amount() > 1) {
+                var i = self.data.startnr();
+                for (i = self.data.startnr(); i < (self.data.startnr() + self.data.amount()); i++) {
+                    name = self.data.name() + '-' + i;
```
kvanhijf, line 81: var name

webapps/frontend/app/viewmodels/wizards/createfromtemplatewizard/gather.js
```diff
@@ -155,6 +176,16 @@ define([
                 self.data.vm().load();
                 self.data.selectedPMachines([]);
             }
+            generic.xhrAbort(self.loadVMachinesHandle);
+            self.loadVMachinesHandle = api.get('vmachines/')
```
khenderick: You could add a `contents = 'name'` here so you only have to do one call instead of possibly hundreds. That would take that long that the check doesn't make much sense (you most likely already completed the wizard before the calls are completed)

kvanhijf: You're not waiting for the vms to be loaded, so the validation you do when clicking next might be incorrect

webapps/frontend/app/viewmodels/wizards/createfromtemplatewizard/gather.js
khenderick, line 81: Use `var name, i = self.data.startnr();` before the for loop, then you don't initialize a new variable every iteration.

webapps/frontend/app/viewmodels/wizards/createfromtemplatewizard/gather.js
```diff
@@ -66,6 +67,28 @@ define([
...
+            if (self.data.amount() === 1) {
+                if (self.data.vMachinesNames().contains(self.data.name())) {
+                    valid = false;
+                    fields.push('name');
+                    reasons.push($.t('ovs:wizards.createft.gather.duplicatename', {
+                        name: self.data.name()
+                    }));
+                }
+            }
+            else {
```
khenderick: Should be `} else {`.

webapps/frontend/app/viewmodels/wizards/createfromtemplatewizard/gather.js
```diff
@@ -155,6 +178,13 @@ define([
                 self.data.vm().load();
                 self.data.selectedPMachines([]);
             }
+            generic.xhrAbort(self.loadVMachinesHandle);
+            self.loadVMachinesHandle = api.get('vmachines/', { queryparams: {contents: 'name'}})
+                .done(function(data) {
+                    $.each(data.data, function(index, item) {
+                        self.data.vMachinesNames.push(item.name);
```
khenderick, line 184: I'd expect that somewhere `self.data.vMachineNames` would be reset to an empty list (typically in index.js), because now this list will grow larger with every time the wizard is opened.

PR #472, opened 2016-03-18
OVS-4365 Fix disk sync

eduardBM:
- correctly identify partition name and number

### DIFF
--- a/ovs/lib/storagerouter.py
+++ b/ovs/lib/storagerouter.py
@@ -114,27 +114,18 @@ class StorageRouterController(object):
                             used_size, _ = client.run('du -B 1M -d 0 {0}'.format(storagedriver_partition.path)).split('\t')
                             directory_used_size = int(used_size)
                         except Exception as ex:
-                            logger.error('Failed to get directory usage for {0}. {1}'.format(storagedriver_partition.path, ex))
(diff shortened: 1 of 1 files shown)


ovs/lib/storagerouter.py
```diff
@@ -114,27 +114,18 @@ def get_metadata(storagerouter_guid):
...
-                            logger.error('Failed to get directory usage for {0}. {1}'.format(storagedriver_partition.path, ex))
+                            logger.warning('Failed to get directory usage for {0}. {1}'.format(storagedriver_partition.path, ex))
                     used_space_by_roles += directory_used_size
 
                 partition_available_space = None
                 if disk_partition.mountpoint is not None:
-                    disk_partition_device = '/dev/{0}{1}'.format(disk.name, disk_partition.id)
+                    disk_partition_device = os.readlink(disk_partition.path).replace('../..', '/dev')
                     try:
                         output = client.run("df -Pl | grep {0}".format(disk_partition_device))
```
kvanhijf: Please use the -B option to retrieve the values in size required (Kb, Mb, Gb) and --output to only list the fields you require
Also i wouldn't do this in a try except
Check if disk_partition_device exists and execute command if it does

eduardBM: I put it in a try/except because of all the conversion, line splitting, etc.
Changed the command to return required value with less parsing.

PR #475, opened 2016-03-18
OVS-4253 Celery inspect ping - add timeout

eduardBM:


khenderick:
Please merge from eugene-updates to master. Don't make the same commit on both.

PR #483, opened 2016-03-23
OVS-3682 Let SSHClient connect at __init__

eduardBM:


ovs/extensions/generic/sshclient.py
```diff
@@ -142,6 +142,7 @@ def __init__(self, endpoint, username='ovs', password=None):
                 client.is_connected = types.MethodType(is_connected, client)
                 SSHClient.client_cache[key] = client
             self.client = SSHClient.client_cache[key]
+        self._connect()
```
kvanhijf, line 145: Since we connect in the initialization of the SSHClient class, the @connected decorator can be removed (i guess)

PR #491, opened 2016-04-01
OVS-4415 - Enable heartbeats

eduardBM:


ovs/lib/setup.py
```diff
@@ -473,6 +473,11 @@ def setup_node(ip=None, force_type=None):
                                 raise
 
             print ''
+            print Interactive.boxed_message(['Open vStorage securely sends a minimal set of error, usage and health',
```
kvanhijf: This message should only be printed when 'enable_heartbeats' is True
Can be False when using the preconfig

PR #492, opened 2016-04-05
OVS-4350 - Get service status

eduardBM:
- fix for workers in pre/start during setup
- service manager returns status correctly


ovs/extensions/services/upstart.py
```diff
@@ -96,7 +96,7 @@ def get_service_status(name, client, return_output=False):
                     return status, output
                 return status
             # Normal cases - or if the above code didn't yield an outcome
-            if 'start' in output or 'is running' in output:
+            if 'start/running' in output or 'is running' in output:
                 if return_output is True:
                     return True, output
                 return True
```
kvanhijf, line 102: Shouldn't stop also be changed to stop/waiting ?
I don't know if there some sort of pre-stop for stopping a service

ovs/lib/helpers/toolbox.py
```diff
@@ -197,3 +197,18 @@ def change_service_state(client, name, state, logger=None):
         else:
             logger.debug('  {0:<15} - Service {1} {2}'.format(client.ip, name, action))
             print '  [{0}] {1} {2}'.format(client.ip, name, action)
+
+    @staticmethod
+    def wait_for_service(client, name):
```
kvanhijf: I'd add a status to this call
Default can be True, so you wait until the service is running
But this option also allows you to wait until a service is stopped entirely

ovs/extensions/services/upstart.py
eduardBM, line 102: Stop is always stopped.

ovs/lib/helpers/toolbox.py
```diff
@@ -197,3 +197,18 @@ def change_service_state(client, name, state, logger=None):
...
+        while tries > 0:
+            status = ServiceManager.get_service_status(name, client)
+            if status is True:
+                break
+            print('... waiting for workers')
+            tries -= 1
+            time.sleep(10 - tries)
+        status, output = ServiceManager.get_service_status(name, client, True)
+        if status is False:
+            raise RuntimeError('Service {0} is not running: {1}'.format(name, output))
```
kvanhijf: No newline at end of file

ovs/lib/helpers/toolbox.py
kvanhijf: Name parameters like all other functions in this file

ovs/lib/helpers/toolbox.py
```diff
@@ -197,3 +197,24 @@ def change_service_state(client, name, state, logger=None):
...
             logger.debug('  {0:<15} - Service {1} {2}'.format(client.ip, name, action))
             print '  [{0}] {1} {2}'.format(client.ip, name, action)
+
+    @staticmethod
+    def wait_for_service(client, name, status=True):
+        """
+        Wait for service to enter status
+        :param client: SSHClient to run commands
+        :param name: name of service
+        :params status: True - running/False - not running
```
kvanhijf: :params should be :param

ovs/lib/helpers/toolbox.py
```diff
@@ -197,3 +197,24 @@ def change_service_state(client, name, state, logger=None):
...
+        """
+        ServiceManager.enable_service(name, client=client)
+        tries = 10
+        while tries > 0:
+            service_status = ServiceManager.get_service_status(name, client)
+            if service_status == status :
```
kvanhijf: space between status and the colon
`status :`

ovs/lib/helpers/toolbox.py
```diff
@@ -197,3 +197,24 @@ def change_service_state(client, name, state, logger=None):
...
+        :param name: name of service
+        :param status: True - running/False - not running
+        """
+        ServiceManager.enable_service(name, client=client)
+        tries = 10
+        while tries > 0:
+            service_status = ServiceManager.get_service_status(name, client)
+            if service_status == status:
+                break
+            print('... waiting for service')
```
khenderick, line 215: I assume it's too late to get rid of this print and replace it by logger.whatever like we do at every other place?

ovs/lib/helpers/toolbox.py
```diff
@@ -197,3 +197,24 @@ def change_service_state(client, name, state, logger=None):
...
+
+    @staticmethod
+    def wait_for_service(client, name, status=True):
+        """
+        Wait for service to enter status
+        :param client: SSHClient to run commands
+        :param name: name of service
+        :param status: True - running/False - not running
+        """
+        ServiceManager.enable_service(name, client=client)
```
khenderick, line 209: So when somebody actually wants to wait for a service to be in a given state, we enable it. So if somebody just disabled a service, stopped is and now wants to wait for it to be actually stopped, we just enabled the service again.

PR #493, opened 2016-04-06
OVS-4350 - Get service status

eduardBM:
- fix issues in Toolbox.wait_for_service

### DIFF
--- a/ovs/lib/setup.py
+++ b/ovs/lib/setup.py
@@ -446,7 +446,6 @@ class SetupController(object):
                         if configure_rabbitmq is None:
                             configure_rabbitmq = r_config.get('setup', 'configure_rabbitmq')
 
-
                         print 'Analyzing cluster layout'
                         logger.info('Analyzing cluster layout')
                         config = ArakoonClusterConfig('ovsdb')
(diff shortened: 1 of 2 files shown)


ovs/lib/setup.py
```diff
@@ -1228,7 +1228,8 @@ def _setup_extra_node(cluster_ip, master_ip, cluster_name, unique_id, ip_client_
 
         logger.debug('Check ovs-workers')
         # Workers are started by ovs-watcher-framework, but for a short time they are in pre-start
-        Toolbox.wait_for_service(target_client, 'workers')
+        ServiceManager.enable_service('workers', client=target_client)
```
kvanhijf, line 1231: Eduard, i don't know how you usually work, but you really need to make changes to your way of working or have some basic guidelines for yourself noted down somewhere, because every time you forget something
In this case, there is another occurrence of 'wait_for_service' which you didnt adjust to your changes -->  3 remarks
1. You didnt test, because ovs setup would have always failed
2. I even remember you implemented this new method 2 times, why didnt you remember?
3. My PyCharm shows this error on the scrollbar mentioning, not all arguments are covered, why doesn't this do this for you? (after mentioning several times)

PR #494, opened 2016-04-07
OVS-4419 - Include both fqdn and short hostname in /etc/hosts

eduardBM:


ovs/lib/setup.py
```diff
@@ -772,13 +772,19 @@ def _prepare_node(cluster_ip, nodes, known_passwords, ip_client_map, hypervisor_
                 authorized_keys += '{0}\n'.format(root_pub_key)
             if ovs_pub_key not in authorized_keys:
                 authorized_keys += '{0}\n'.format(ovs_pub_key)
+
+            node_fqdn = node_client.run('hostname -f')
+            mapping[node] = [node_fqdn]
+            all_hostnames.add(node_fqdn)
+
             node_hostname = node_client.run('hostname')
```
pploegaert: I'd use hostname -s here, so we have both (short / long) and in any case they would be identical, they would be skipped as part of your code

PR #500, opened 2016-04-11
OVS-4431 - Implement new list_volumes call + Various other speed tweaks

khenderick:


ovs/dal/helpers.py
```diff
@@ -332,3 +333,29 @@ def execute(function, start, end):
...
+class timer(object):
+    """
+    Can be used for timing pieces of code
+    """
+
+    def __init__(self, name, ms=False):
+        """
+        Creates a timer
+        :param name: The name of the timer, will be printed
+        :param ms: Indicates whether it should be printed in seconds or miliseconds
```
kvanhijf: It should be `milliseconds`
Add types too please

ovs/lib/vdisk.py
```diff
@@ -186,6 +186,9 @@ def resize_from_voldrv(volumename, volumesize, volumepath, storagedriver_id):
             disk.volume_id = volumename
             disk.size = volumesize
```
kvanhijf, line 187: disk.size is filled out twice here

PR #508, opened 2016-04-14
ovs-4472-task-ready-status-race-condition

pploegaert:


webapps/api/backend/views/tasks.py
```diff
@@ -59,15 +59,16 @@ def retrieve(self, pk):
         :param pk: Primary key
         """
         result = celery.AsyncResult(pk)
+        status = result.status
         if result.successful():
```
kvanhijf: Perhaps also pre-populate the result.successful, similar to the result.status

ovs/extensions/api/client.py
```diff
@@ -170,10 +170,10 @@ def wait_for_task(self, task_id, timeout=None):
         """
         start = time.time()
         task_metadata = {'ready': False}
-        while task_metadata['ready'] is False:
+        while task_metadata['status'] not in ('FAILURE', 'SUCCESS'):
```
khenderick: The `status` key is not in `task_metadata` at this point. But because of the duplicate `... not in ('FAILURE', 'SUCCESS')`, it might be nicer to have something like

```
finished = False
while finished is False:
    # do the actual call
    finished = task_metadata['status'] in ['FAILURE', 'SUCCESS']
    if finished is False:
        time.sleep(1)
```

webapps/api/backend/views/tasks.py
```diff
@@ -59,15 +59,16 @@ def retrieve(self, pk):
         :param pk: Primary key
         """
         result = celery.AsyncResult(pk)
+        status = result.status
```
khenderick: You cannot name your variable `status` here, as this shadows the rest_framework status module, which will cause some error that `status` has no property `HTTP_200_OK`.

PR #517, opened 2016-04-25
Update etcd.md

saelbrec:


docs/etcd.md
```diff
@@ -1 +1,269 @@
...
+
+The ```$host_id``` in above keys can be found on the respective node in ```/etc/openvstorage_id```
+
+#### Arakoon
+All arakoon keys stated below are relative to ```/ovs/arakoon```. E.g. ```ovsdb``` will be ```/ovs/arakoon/ovsdb```.
+
+##### FWK
+```
+$clustername/metadata = {"internal": True|False,
+                         "type": "FWK",
```
khenderick: I'd put all options here (like `FWK|SD|NSM|ABM`)

docs/etcd.md
```diff
@@ -1 +1,269 @@
...
+                         "in_use": True|False}
+$clustername/config = "[global]
```
khenderick: I think it would be better to define the contents like `...config = "<raw arakoon configuration format>"` or something similar, optionally referring to the arakoon documentation. Otherwise people can mis-interpret this as being a guide for the arakoon configuration format.

PR #521, opened 2016-04-26
#125 - Make accelerated backend configurable per Storage Driver

kvanhijf:


ovs/extensions/api/client.py
```diff
@@ -15,13 +15,20 @@
...
-import urllib
-import base64
 import time
+import base64
+import urllib
+import hashlib
 import requests
 
 
+class ForbiddenException(Exception):
```
khenderick: Should inherit `RuntimeError` for backwards compatibility.

ovs/extensions/api/client.py
```diff
@@ -106,18 +127,27 @@ def _process(self, response):
         except:
             raise RuntimeError('Could not parse returned data: {0}: {1}'.format(response.status_code, response.text))
 
+    def _call(self, api, params, function, **kwargs):
+        if not api.endswith('/'):
+            api = '{0}/'.format(api)
+        if not api.startswith('/'):
+            api = '/{0}'.format(api)
+        self._token = self._volatile_client.get(self._key)
```
khenderick: You should check whether `self._volatile_client` is not `None`.

ovs/extensions/api/client.py
```diff
@@ -106,18 +127,27 @@ def _process(self, response):
...
+        headers, url = self._prepare(params=params)
+        try:
+            return self._process(function(url=url.format(api), headers=headers, verify=self._verify, **kwargs))
+        except ForbiddenException:
+            self._token = None
```
khenderick, line 144: Before setting it to `None`, you can check whether it already contains `None`. If that's the case, you can just raise this exception immediately.

khenderick:
Please make sure to - at least - add unittests for the `OVSClient`.

PR #523, opened 2016-04-28
OVS-4477 Itersafe race condition

eduardBM:
- catch KeyNotFoundException

### DIFF
--- a/ovs/dal/datalist.py
+++ b/ovs/dal/datalist.py
@@ -22,7 +22,7 @@ import random
 import hashlib
 from random import randint
 from ovs.dal.helpers import Descriptor, HybridRunner
-from ovs.dal.exceptions import ObjectNotFoundException
+from ovs.dal.exceptions import ObjectNotFoundException, RaceConditionException
 from ovs.extensions.storage.exceptions import KeyNotFoundException
 from ovs.extensions.storage.volatilefactory import VolatileFactory
 from ovs.extensions.storage.persistentfactory import PersistentFactory
@@ -206,6 +206,7 @@ class DataList(object):
             keys = ['{0}{1}'.format(prefix, guid) for guid in self._guids]
             successful = False
             tries = 0
+            entries = []
             while successful is False:
                 tries += 1
                 if tries > 5:
@@ -277,6 +278,7 @@ class DataList(object):
             keys = ['{0}{1}'.format(prefix, guid) for guid in self._guids]
             successful = False
             tries = 0
+            entries = []
             while successful is False:
                 tries += 1
                 if tries > 5:
--- a/ovs/dal/datalist.py
+++ b/ovs/dal/datalist.py
@@ -203,12 +203,20 @@ class DataList(object):
         prefix = '{0}_{1}_'.format(DataObject.NAMESPACE, object_type_name)
 
         if self._guids is not None:
-            try:
-                entries = list(self._persistent.get_multi(['{0}{1}'.format(prefix, guid) for guid in self._guids]))
-            except KeyNotFoundException as knfe:
-                guid_not_found = knfe.message.split('_')[-1]
-                self._guids.remove(guid_not_found)
-                entries = list(self._persistent.get_multi(['{0}{1}'.format(prefix, guid) for guid in self._guids]))
+            keys = ['{0}{1}'.format(prefix, guid) for guid in self._guids]
+            successfull = False
+            tries = 0
+            while successfull is False:
+                tries += 1
+                if tries > 5:
+                    raise RaceConditionException()
+                try:
+                    entries = list(self._persistent.get_multi(keys))
+                    successfull = True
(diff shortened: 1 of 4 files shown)


ovs/dal/dataobject.py
```diff
@@ -679,12 +679,15 @@ def delete(self, abandon=None, _hook=None):
...
-                                    except ObjectNotFoundException:
-                                        pass
+                                try:
+                                    for item in items.itersafe():
+                                        setattr(item, info['key'], None)
+                                        try:
+                                            item.save()
+                                        except ObjectNotFoundException:
+                                            pass
+                                except KeyNotFoundException:
```
khenderick: The issue is not only at/in the itersafe, but based on the reported stacktrace it's caused by the multi-get that got a non-existing key. It can basically happen at any place where a query is executed. The query itself should be smart enough to re-execute without the missing key.

ovs/dal/dataobject.py
```diff
@@ -736,7 +739,7 @@ def delete(self, abandon=None, _hook=None):
                 self._persistent.apply_transaction(transaction)
                 successful = True
             except KeyNotFoundException as ex:
-                if ex.message != self._key:
+                if self._key not in ex.message:
```
khenderick: Hmm, odd, I was under the impression that ex.message is always exactly equal to the key that was not found. Can you add the example on which you based this change in a comment?

eduardBM: `KeyNotFoundException: 'ovs_data_vdisk_55e15197-183f-43dd-b25c-ef2425d6241e'

In [4]: %debug

> /opt/OpenvStorage/ovs/extensions/storage/persistent/pyrakoonstore.py(224)apply_transaction()
>     223         except ArakoonNotFound as field:
> --> 224             raise KeyNotFoundException(field)
>     225 

ipdb> up

> /opt/OpenvStorage/ovs/dal/dataobject.py(740)delete()
>     739                 if ex.message != self._key:
> --> 740                     raise
>     741                 successful = True

ipdb> ex.message
ArakoonNotFound('ovs_data_vdisk_55e15197-183f-43dd-b25c-ef2425d6241e',)
ipdb> self._key
'ovs_data_vdisk_55e15197-183f-43dd-b25c-ef2425d6241e'
ipdb> 
`

khenderick: Aha, interesting, that was not on purpose. It might be a better idea though to change the pyrakoonstore to pass the field as message instead of the exception. `field` in there is the exception and not the field itself, so it's name is wrong.

ovs/dal/datalist.py
```diff
@@ -203,7 +203,13 @@ def _execute_query(self):
         prefix = '{0}_{1}_'.format(DataObject.NAMESPACE, object_type_name)
 
         if self._guids is not None:
-            entries = list(self._persistent.get_multi(['{0}{1}'.format(prefix, guid) for guid in self._guids]))
+            entries = []
+            for guid in self._guids[:]:
+                try:
+                    entries.append(self._persistent.get('{0}{1}'.format(prefix, guid)))
```
khenderick: That will introduce an enormous slowdown of the code in case of a missing key. I think a better approach might be to catch the `KeyNotFoundException` and remove the reported field from the list, after which you can execute a multi-get again. Maybe you can for example do a maximum of 5 tries (we also retry 5 times in other places when we suffer from race conditions).

It's probably a good idea to first create a separate `guids` or sth like that, so you can easily remove entries from it.

ovs/dal/datalist.py
```diff
@@ -203,7 +203,13 @@ def _execute_query(self):
         prefix = '{0}_{1}_'.format(DataObject.NAMESPACE, object_type_name)
 
         if self._guids is not None:
-            entries = list(self._persistent.get_multi(['{0}{1}'.format(prefix, guid) for guid in self._guids]))
+            try:
+                entries = list(self._persistent.get_multi(['{0}{1}'.format(prefix, guid) for guid in self._guids]))
```
khenderick: Can you build a list of the keys first? Something like:

```
keys = ['{0}{1}'.format(prefix, guid) for guid in self._guids]
```

Then, if an error occurs, you can remove the key immediately from that list, and then parse out the guid to remove it from _guids. This will make sure you don't have loop through all guids every time you retry.

Also, like stated in the previous comment, please re-try this 5 times (analog code like in `DataObject.delete()`. And it might be good to in that case raise a `RaceConditionException`, just like the delete mentioned earlier.

ovs/dal/datalist.py
```diff
@@ -203,7 +203,13 @@ def _execute_query(self):
...
+            except KeyNotFoundException as knfe:
+                guid_not_found = knfe.message.split('_')[-1]
```
khenderick: You should not use splits here, but instead make use of the `prefix` variable.

PR #524, opened 2016-05-02
#125 - Introduced unittest helper class

kvanhijf:
Removed some obsolete arakoon files


ovs/dal/hybrids/storagedriver.py
```diff
@@ -24,14 +24,14 @@
...
 
-logger = LogHandler.get('dal', name='hybrid')
-
 
 class StorageDriver(DataObject):
     """
     The StorageDriver class represents a Storage Driver. A Storage Driver is an application
     on a Storage Router to which the vDisks connect. The Storage Driver is the gateway to the Storage Backend.
     """
+    _logger = LogHandler.get('dal', name='hybrid')
```
khenderick, line 33: I wonder whether a hybrid can use the logger available in the `DataObject`. If not, try by - in `DataObject` - assign the logger to sth like `self._logger`, and then I think you can just use `self._logger` in the hybrids as well.

ovs/dal/performance_test.py
```diff
@@ -28,17 +27,20 @@
...
 
 
-class LotsOfObjects(TestCase):
+class LotsOfObjects(object):
     """
     Executes a performance test by working with a large set of objects
     """
+    def __init__(self):
+        """
+        Dummy init method
```
khenderick: It's not dummy anymore

ovs/dal/hybrids/storagedriver.py
kvanhijf, line 33: Keeping it as it is after discussing the issue

ovs/dal/performance_test.py
kvanhijf: Fixed

PR #525, opened 2016-05-03
OVS-4342 - Gather scrubwork is not properly exited after run

eduardBM:
- make sure remote tasks are able to run (workers are running)
- timeout in case workers go down while executing task (1 hr)


ovs/lib/helpers/celery_toolbox.py
```diff
@@ -0,0 +1,81 @@
...
+
+import datetime
+import time
+from celery.task.control import revoke
+from ovs.dal.lists.storagerouterlist import StorageRouterList
+from ovs.extensions.generic.sshclient import SSHClient
+from ovs.extensions.services.service import ServiceManager
+from ovs.log.logHandler import LogHandler
+
+logger = LogHandler.get('lib', name='celery toolbox')
```
kvanhijf: Please define the `logger` instance inside the method using it
We're trying to minimize the variable declarations outside of `main` and/or classes

ovs/lib/helpers/celery_toolbox.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def manage_running_tasks(tasklist, timesleep=10):
+    """
+    Manage a list of running celery task
+    - discard PENDING tasks after a certain timeout
+    - validate RUNNING tasks are actually running
+    @param tasklist: dict {IP address: AsyncResult}
```
kvanhijf: Notation we use:
`:param tasklist:  Info about tasklist
:type tasklist: dict`

ovs/lib/helpers/celery_toolbox.py
```diff
@@ -0,0 +1,81 @@
...
+    @param timesleep: int (sleep between checks -
+      -for long running tasks it's better to sleep for a longer period of time to reduce number of ssh calls
+    @return: list of results
```
kvanhijf: Notation we use:
`:return: Taskresults
:rtype: list`

ovs/lib/helpers/celery_toolbox.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    ssh_clients = {}
+    tasks_pending = {}
+    tasks_pending_timeout = 10 #800  # 30 minutes
```
kvanhijf: 2 blank spaces between end of line and start of comment and please explain what the 800 means

ovs/lib/helpers/celery_toolbox.py
```diff
@@ -0,0 +1,81 @@
...
+    results = []
+    failed_nodes = []
+    while len(tasklist.keys()) > 0:
+        for ip, task in tasklist.items():
+            if task.state in ('SUCCESS', 'FAILURE'):
+                logger.info('Task {0} finished: {1}'.format(task.id, task.state))
+                results.append(task.get(propagate=False))
+                del tasklist[ip]
```
kvanhijf: Its never a smart idea to delete items from the list or dict you are looping over
However since you use `tasklist.items()` the entire dict is read into memory and its possible to delete items from it while looping over it. If we ever change to `iteritems()`, the code will break. So potentially dangerous

ovs/lib/helpers/celery_toolbox.py
```diff
@@ -0,0 +1,81 @@
...
+                client = ssh_clients[ip]
+                if ServiceManager.get_service_status('workers', client) is False:
+                    logger.error('Service ovs-workers on node {0} appears halted while there is a task PENDING for it {1}. Task will be revoked.'.format(ip, task.id))
+                    revoke(task.id)
+                    del tasklist[ip]
+                    failed_nodes.append(ip)
+                else:
+                    ping_result = task.app.control.inspect().ping()
+                    storage_router = StorageRouterList.get_by_ip(ip)
+                    if "celery@{0}".format(storage_router.name) not in ping_result:
```
kvanhijf: does this also work if the storagerouter is part of a domain and `hostname -f != hostname -s`
Because we store the short hostname in the model for each storagerouter, but celery might use the fqdn (not sure)

ovs/lib/scheduledtask.py
```diff
@@ -37,7 +36,9 @@
 from ovs.extensions.generic.sshclient import SSHClient
 from ovs.extensions.generic.sshclient import UnableToConnectException
 from ovs.extensions.generic.system import System
+from ovs.extensions.services.service import ServiceManager
 from ovs.lib.helpers.decorators import ensure_single
+from ovs.lib.helpers.celery_toolbox import manage_running_tasks
```
kvanhijf: Sort alphabetically plz

ovs/lib/scheduledtask.py
```diff
@@ -225,7 +226,11 @@ def gather_scrub_work():
                     logger.info('Gather Scrub - Storage Router {0:<15} has SCRUB partition at {1}'.format(storage_driver.storagerouter.ip, partition.path))
                     if storage_driver.storagerouter not in scrub_locations:
                         try:
-                            _ = SSHClient(storage_driver.storagerouter)
+                            sshclient = SSHClient(storage_driver.storagerouter)
+                            # Use ServiceManager(sshclient) to make sure ovs-workers are actually running
+                            if ServiceManager.get_service_status('workers', sshclient) is False:
+                                logger.warning('Gather Scrub - Storage Router {0:<15} - workers are not running'.format(storage_driver.storagerouter.ip))
+                            #else:
```
kvanhijf: Why add the storagerouter to the list of `scrub_locations` even when the workers are not running?   And if this is what we want, please remove the `#else:`

ovs/lib/helpers/celery_toolbox.py
```diff
@@ -0,0 +1,81 @@
...
+from celery.task.control import revoke
+from ovs.dal.lists.storagerouterlist import StorageRouterList
+from ovs.extensions.generic.sshclient import SSHClient
+from ovs.extensions.services.service import ServiceManager
+from ovs.log.logHandler import LogHandler
+
+logger = LogHandler.get('lib', name='celery toolbox')
+
+
+def manage_running_tasks(tasklist, timesleep=10):
```
kvanhijf: Use a class instead
All our libraries are classes with static methods only

ovs/lib/scheduledtask.py
```diff
@@ -263,11 +269,10 @@ def gather_scrub_work():
                 local_scrub_location = scrub_info[1]
```
kvanhijf, line 268: The local started scrub task should be handled the same way you process the remote tasks with these changes in place, because the local celery task can also remain in PENDING state if the local workers are stopped for whatever reason
Your `result_set` does not contain this local storagerouter

ovs/lib/scheduledtask.py
```diff
@@ -277,13 +282,50 @@ def gather_scrub_work():
...
                 logger.error('Gather Scrub - Storage Router {0:<15} - Scrubbing failed with error:\n - {1}'.format(local_storage_router.ip, ex))
-        all_results = result_set.join(propagate=False)  # Propagate False makes sure all jobs are waited for even when 1 or more jobs fail
+
+        all_results, failed_nodes = manage_running_tasks(result_set, timesleep=60)  # Check every 60 seconds if tasks are still running
+
+        for index, result in enumerate(all_results):
+            if isinstance(result, list):
+                processed_guids.extend(result)
+            else:
+                logger.error('Gather Scrub - Storage Router {0:<15} - Scrubbing failed with error:\n - {1}'.format(storage_router_list[index].ip, result))
```
kvanhijf: Are we sure the order of `storage_router_list` matches the order in which the `manage_running_tasks` returns the results for each storagerouter?? Because `result_set` which is passed to the function is a dict, which is obviously unordered

ovs/lib/scheduledtask.py
```diff
@@ -247,8 +252,9 @@ def gather_scrub_work():
         local_storage_router = None
         local_scrub_location = None
         local_vdisks_to_scrub = []
-        result_set = ResultSet([])
+        result_set = {}
```
kvanhijf, line 254: Can't we merge these 3 items (`result_set, storage_router_list, scrub_map`
They all contain info about storagerouters
Make it into 1 list with each element being a dict or another list

ovs/lib/scheduledtask.py
```diff
@@ -277,13 +282,50 @@ def gather_scrub_work():
...
+            else:
+                logger.error('Gather Scrub - Storage Router {0:<15} - Scrubbing failed with error:\n - {1}'.format(storage_router_list[index].ip, result))
+
+        result_set = {}
+        for failed_node in failed_nodes:
+            logger.warning('Scrubbing failed on node {0}. Will reschedule on another node.'.format(failed_node))
+            vdisk_guids_to_scrub = scrub_map[failed_node]
+            rescheduled_work = False
+            for storage_router, scrub_location in scrub_locations.items():
+                if storage_router.ip != failed_node:
```
kvanhijf: If you have multiple failed nodes, this code can potentially start scrubbing on a different failed node, than the 1 being checked currently. To prevent this, first do a cleanup of `scrub_locations` by removing all `failed_nodes` from it
Then check if any scrub locations are left and raise if not

ovs/lib/helpers/celery_toolbox.py
eduardBM: That's how it is defined everywhere, to be usable by multiple methods.

ovs/lib/helpers/celery_toolbox.py
eduardBM: ok

ovs/lib/helpers/celery_toolbox.py
eduardBM: ok

ovs/lib/helpers/celery_toolbox.py
eduardBM: my mistake, should have been 1800  # 30 minutes

ovs/lib/helpers/celery_toolbox.py
eduardBM: That's why i used items() instead of iteritems().

ovs/lib/helpers/celery_toolbox.py
eduardBM: Not sure whether celery uses a hostname different from storage_router.name, will check.

ovs/lib/scheduledtask.py
eduardBM: ok

ovs/lib/scheduledtask.py
eduardBM: my mistake, seems i commited the test version of the file, not the final

ovs/lib/helpers/celery_toolbox.py
eduardBM: Ok

ovs/lib/scheduledtask.py
eduardBM, line 268: Local scrub is not executed in another task, it's executed in the same task by the running worker so it can't remain in PENDING.

ovs/lib/scheduledtask.py
eduardBM: Will check

ovs/lib/scheduledtask.py
eduardBM, line 254: result_set is the list of results from each running task (the guids of scrubbed disks) - no info about storagerouters
storage_router_list is the list of storagerouters
scrub_map is a helper dictionary to keep track of what guids were assigned to each storagerouter

so i don't think they can be merged

ovs/lib/scheduledtask.py
eduardBM: Indeed, this line should be "if storage_router.ip not in failed_nodes... "

PR #528, opened 2016-05-17
OVS-4394 - OVS setup improvements

eduardBM:
- fixed issues with rollback

### DIFF
--- a/ovs/lib/setup.py
+++ b/ovs/lib/setup.py
@@ -1064,7 +1064,7 @@ class SetupController(object):
             except ValueError:
                 metadata = None
             if metadata is not None and metadata.internal is True:
-                services.append('arakoon-ovsdb')
+                services.append('arakoon-{0}'.format(cluster_name))
             for service in services:
                 if ServiceManager.has_service(service, client=target_client):
                     ServiceManager.disable_service(service, client=target_client)
--- a/ovs/lib/setup.py
+++ b/ovs/lib/setup.py
@@ -1052,7 +1052,7 @@ class SetupController(object):
                         pass
 
             try:
-                EtcdConfiguration.delete(key='/ovs/framework/{0}'.format('hosts/{0}'.format(machine_id)))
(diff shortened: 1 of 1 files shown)


ovs/lib/setup.py
```diff
@@ -1056,24 +1027,73 @@ def _rollback_setup(target_client, first_node):
...
+            if first_node is True:
+                for key in EtcdConfiguration.base_config.keys() + ['install_time', 'plugins']:
+                    try:
+                        EtcdConfiguration.delete(key='/ovs/framework/{0}'.format(key))
+                    except (EtcdKeyNotFound, EtcdConnectionFailed, KeyError):
+                        pass
 
-            for key in EtcdConfiguration.base_config.keys() + ['install_time', 'plugins', 'hosts/{0}'.format(machine_id)]:
+            try:
+                EtcdConfiguration.delete(key='/ovs/framework/{0}'.format('hosts/{0}'.format(machine_id)))
```
khenderick: Can we change this to `...key='/ovs/framework/hosts/{0}'.format(machine_id)...`?

ovs/lib/setup.py
```diff
@@ -1056,24 +1027,73 @@ def _rollback_setup(target_client, first_node):
...
+            except (EtcdKeyNotFound, EtcdConnectionFailed, KeyError):
+                pass
+
+            #  Arakoon and ETCD must be the last services to be removed
+            for service in ['memcached', 'arakoon-ovsdb']:
```
khenderick: It's possible that the `arakoon-ovsdb` service won't exist if an external arakoon cluster was used.

eduardBM: It is checked on next line. (if ServiceManager.has_service...)

ovs/lib/setup.py
eduardBM: Indeed...

ovs/lib/setup.py
khenderick: Yeah, but I mean that it can have a different name. See a bunch of lines below on how to retrieve the correct name.

PR #537, opened 2016-05-20
OVS-4525 - Framework marks living volumedriver as offline

eduardBM:
- fix check for volumedriver

### DIFF
--- a/ovs/lib/storagedriver.py
+++ b/ovs/lib/storagedriver.py
@@ -34,13 +34,13 @@ from ovs.dal.lists.vdisklist import VDiskList
 from ovs.extensions.db.arakoon.ArakoonInstaller import ArakoonInstaller, ArakoonClusterConfig
 from ovs.extensions.db.etcd.configuration import EtcdConfiguration
 from ovs.extensions.generic.remote import remote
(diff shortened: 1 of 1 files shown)


ovs/lib/storagedriver.py
```diff
@@ -84,26 +84,35 @@ def update_status(storagedriver_id):
...
+                StorageDriverController._logger.error('SSHClient connectivity check failed, assuming host {0} is halted. {1}'.format(pmachine.name, ex))
                 host_status = 'HALTED'
-            if host_status != 'RUNNING':
-                # Host is stopped
-                storagedriver_client = StorageDriverClient.load(storagedriver.vpool)
-                storagedriver_client.mark_node_offline(str(storagedriver.storagedriver_id))
+            else:
+                try:
+                    with remote(client.ip, [LocalStorageRouterClient]) as rem:
+                        path = 'etcd://127.0.0.1:2379/ovs/vpools/{0}/hosts/{1}/config'.format(storagedriver.vpool.guid,
```
khenderick: Can you create a `StorageDriverConfiguration` object (that can load the config based on vpool etc), and then use `.remote_path` property. If we have to move the config location around, we only have to update that one.

PR #542, opened 2016-05-25
OVS-4474 - Vpool overview imports containers from plugin

eduardBM:
- make overview work if plugin is not present


webapps/frontend/app/viewmodels/wizards/addvpool/data.js
```diff
@@ -157,20 +157,20 @@ define(['knockout'], function(ko){
                 wizardData.albaPreset(undefined);
                 return []
             }
-            if (!wizardData.albaBackend().enhancedPresets().contains(wizardData.albaPreset())){
-                wizardData.albaPreset(wizardData.albaBackend().enhancedPresets()[0]);
+            if (wizardData.albaBackend().presets.indexOf(wizardData.albaPreset()) === -1){
```
khenderick: Why was this changed to `indexOf` instead of `contains`?

webapps/frontend/app/viewmodels/wizards/addvpool/gather_vpool.js
```diff
@@ -409,4 +393,4 @@ define([
             }
         };
     };
-});
+});
```
khenderick: Please make sure there's a newline at the end of every file.

Issue #558, opened 2016-05-30
rabbitmq warning reports connection_closed_abruptly due to celery ping heartbeat


jeroenmaelbrancke:
ovs version:

```
"versions": {
            "openvstorage-backend-core": "1.7.0-rev.588.2e534bc-1",
            "alba": "0.9.5",
            "openvstorage-webapps": "2.7.0-rev.3451.9fa0267-1",
            "openvstorage-backend-webapps": "1.7.0-rev.588.2e534bc-1",
            "openvstorage-core": "2.7.0-rev.3451.9fa0267-1",
            "volumedriver-server": "5.7.2-dev.201604211506.042c8a8",
            "openvstorage": "2.7.0-rev.3451.9fa0267-1",
            "openvstorage-cinder-plugin": "1.2.0-rev.26.fd48d23-1",
            "openvstorage-backend": "1.7.0-rev.588.2e534bc-1",
            "volumedriver-base": "5.7.2-dev.201604211506.042c8a8",
            "arakoon": "1.9.3"
}
```

Every minute we received a warning report from rabbitmq:

```
=WARNING REPORT==== 30-May-2016::14:35:07 ===
closing AMQP connection <0.21929.600> (10.100.199.1:55464 -> 10.100.199.1:5672):
connection_closed_abruptly

=WARNING REPORT==== 30-May-2016::14:36:07 ===
closing AMQP connection <0.23050.600> (10.100.199.1:56894 -> 10.100.199.1:5672):
connection_closed_abruptly

=WARNING REPORT==== 30-May-2016::14:37:06 ===
closing AMQP connection <0.24076.600> (10.100.199.1:58346 -> 10.100.199.1:5672):
connection_closed_abruptly
```

the celery inspect ping is the problem (connection to rabbitmq is not properly closed):

```
root      6309  0.0  0.0   4448   780 ?        S    14:02   0:00 /bin/sh -c /usr/bin/celery inspect ping -b amqp://ovs:0penv5tor4ge@10.100.199.1// --timeout=5 2> /dev/null | grep OK | perl -pe 's/?\[[0-9;]*m//g' || true
root      6310  8.4  0.3  92288 29480 ?        S    14:02   0:00 python2.7 /usr/bin/celery inspect ping -b amqp://ovs:0penv5tor4ge@10.100.199.1// --timeout=5
```

[Celery ping](https://github.com/openvstorage/framework/blob/master/ovs/extensions/generic/heartbeat.py#L54)

Solution:

```
[...]

PR #561, opened 2016-05-31
rabbitmq warning reports connection_closed_abruptly due to celery ping heartbeat

jeroenmaelbrancke:
[issue 558](https://github.com/openvstorage/framework/issues/558)

### DIFF
--- a/ovs/extensions/generic/heartbeat.py
+++ b/ovs/extensions/generic/heartbeat.py
@@ -53,8 +53,6 @@ class HeartBeat(object):
         with Celery(broker=amqp) as celery:
             worker_states = celery.control.inspect().ping()
 
-        celery.close()
-
         routers = StorageRouterList.get_storagerouters()
         for node in routers:
             if node.heartbeats is None:
--- a/ovs/extensions/generic/heartbeat.py
+++ b/ovs/extensions/generic/heartbeat.py
@@ -50,8 +50,9 @@ class HeartBeat(object):
                                             EtcdConfiguration.get('/ovs/framework/messagequeue|password'),
                                             EtcdConfiguration.get('/ovs/framework/hosts/{0}/ip'.format(machine_id)))
 
-        celery = Celery(broker=amqp)
-        worker_states = celery.control.inspect().ping()
+        with Celery(broker=amqp) as celery:
+            worker_states = celery.control.inspect().ping()
+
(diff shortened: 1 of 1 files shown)


ovs/extensions/generic/heartbeat.py
```diff
@@ -50,21 +50,23 @@ def pulse():
...
+            if worker_states.get('celery@{0}'.format(node.name), {}).get('ok') == 'pong':
                 node.heartbeats['celery'] = current_time
             if node.machine_id == machine_id:
                 node.heartbeats['process'] = current_time
             else:
                 try:
                     # check timeout of other nodes and clear arp cache
                     if node.heartbeats and 'process' in node.heartbeats:
-                        if current_time - node.heartbeats['process'] >= HeartBeat.ARP_TIMEOUT:
+                        if current_time - node.heartbeats['process'] >= 30:
```
khenderick: Why changing this to `30` instead of using the constant?

ovs/extensions/generic/heartbeat.py
```diff
@@ -50,21 +50,23 @@ def pulse():
                                             EtcdConfiguration.get('/ovs/framework/messagequeue|password'),
                                             EtcdConfiguration.get('/ovs/framework/hosts/{0}/ip'.format(machine_id)))
 
-        celery_path = OSManager.get_path('celery')
-        worker_states = check_output("{0} inspect ping -b {1} --timeout=5 2> /dev/null | grep OK | perl -pe 's/\x1b\[[0-9;]*m//g' || true".format(celery_path, amqp), shell=True)
+        celery = Celery(broker=amqp)
```
khenderick: Maybe using a `with` statement, so the close is always executed.

ovs/extensions/generic/heartbeat.py
```diff
@@ -50,21 +50,23 @@ def pulse():
...
+        worker_states = celery.control.inspect().ping()
```
khenderick: Are we sure this call finished within a reasonable time (especially if the RMQ cluster is borked up)?

jeroenmaelbrancke: Default timeout is 1 sec. [celery.app.control](http://docs.celeryproject.org/en/latest/reference/celery.app.control.html)

Issue #552, opened 2016-05-26
'avahi-daemon --reload' fails during setup


kinvaris:
You reload avahi and then restart it as seen here:
https://github.com/openvstorage/framework/blob/master/ovs/lib/setup.py#L1595
Why not just restart it, because sometimes avahi gives problems after installation:

```
root@ctl01:~# avahi-daemon --reload
Failed to kill daemon: No such file or directory    
root@ctl01:~# service avahi-daemon restart
avahi-daemon stop/waiting
avahi-daemon start/running, process 17484
root@ctl01:~# avahi-daemon --reload
```

wimpers:
@kinvaris please create a PR to restart.

PR #563, opened 2016-05-31
remove avahi reload, instead do avahi restart

kinvaris:
#552


kinvaris:
@khenderick: can you verify and merge this PR?

PR #565, opened 2016-06-01
Compute nodes are using the localhost rabbitmq instead of the cluster.

jeroenmaelbrancke:
Fixed this issue by using the /ovs/framework/messagequeue/endpoints.
Code based on ovs/celery_run.py.
@khenderick could you please do a code review of this?
Did you already test the multi-node string?
Should i make a ticket to get all the rabbitmq ips in the endpoints field?

```
etcdctl get /ovs/framework/messagequeue
{"endpoints": ["10.100.199.1:5672"], "protocol": "amqp", "queues": {"storagedriver": "volumerouter"}, "user": "", "password": "", "metadata": {"internal": true}}
```

### DIFF
--- a/ovs/extensions/generic/heartbeat.py
+++ b/ovs/extensions/generic/heartbeat.py
@@ -45,10 +45,12 @@ class HeartBeat(object):
 
         current_time = int(time.time())
         machine_id = System.get_my_machine_id()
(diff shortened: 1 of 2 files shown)


khenderick:
Caused by https://github.com/openvstorage/framework/blob/master/ovs/lib/setup.py#L1240 introduced by 2b304ca.

jeroenmaelbrancke:
@khenderick could you please do a code review of this?
Changed a couple of things cause celery inspect ping doesn't work with multiple brokers....

PR #568, opened 2016-06-02
OVS-4568 - Etcd default timings update

pploegaert:
work-in-progress - increased timings to determine impact on test results


khenderick:
Whoops, the systemd service files are not yet updated.

PR #586, opened 2016-06-08
Edge specs

wimpers:
@khenderick @saelbrec please review

### DIFF
--- a/specs/Edge.md
+++ b/specs/Edge.md
@@ -12,7 +12,7 @@ The [Edge](https://github.com/openvstorage/home/wiki/Edge) is a lightweight soft
 
 ## Required changes
 - Implement storage and management IP correctly
-    - DTL and MDS should use the management Ip
+    - DTL and MDS should use the storage Ip
     - Edge/NFS should use the storage IP
 -  Storage Router detail page
     - Change IP to Management IP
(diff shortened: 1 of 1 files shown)


specs/Edge.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Design choices
+- No Edge manager is required
+ - This means we can't list all disks connected to certain host without retrieving the whole list
+- If possible don't model the edge
+
+
+## Required changes
+- Implement storage and management IP correctly
+    - DTL and MDS should use the management Ip
```
khenderick: Shouldn't they also use the storage ip?

wimpers: I was under the impression this was mgmt ip? @saelbrec  can you shed  a light?

saelbrec: Best practice is to configure storage ip on the network interface with the lowest latency, which is what you want for DTL & MDS.

wimpers:
Updated the spec. Can you verify if the spec is now correct?

khenderick:
Looks good to me. However, for some items the actual implementation will depend on the available information.

saelbrec:
@redlicha , @cnanakos ,
Please check the specs on what is required in the GUI and then decide on how this can be retrieved.
We discussed this in the past and I thought there was a consensus to be able to retrieve this via the volumedriver XMLRPC API. Most likely included in the volume.info ?
Whatever needs to be done from your end can be done under the umbrella of this story.

PR #610, opened 2016-06-14
Allow a cluster to be configured as "RDMA"

khenderick:


ovs/lib/storagerouter.py
```diff
@@ -835,9 +835,10 @@ def add_vpool(parameters):
             EtcdConfiguration.set(config_tree.format('main'), json.dumps({
                 'log_level': 'info',
                 'port': alba_proxy.service.ports[0],
-                'ips': ['127.0.0.1'],
+                'ips': [storagedriver.storage_ip],
                 'manifest_cache_size': 16 * 1024 * 1024 * 1024,
                 'fragment_cache': fragment_cache_info,
+                'transport': 'rdma' if EtcdConfiguration.get('/ovs/framework/rdma') else 'tcp',
```
kvanhijf: The RDMA check can be combined into 1 variable instead of 3 or 4 calls to Etcd

kvanhijf:
The setup file has a preconfig validation method which should validate the RDMA entry (if present)

PR #620, opened 2016-06-16
Fix some long-standing unittest issues

kvanhijf:
Now they all run in memory by default


webapps/api/tests/test_authentication.py
```diff
@@ -368,26 +400,39 @@ def test_specify_scopes(self):
...
+        self.assertIn('access_token', json.loads(response.content))
         access_token = json.loads(response.content)['access_token']
+
+        time.sleep(180)
         header = 'Bearer {0}'.format(access_token)
         request = self.factory.get('/', HTTP_X_REAL_IP='127.0.0.1', HTTP_AUTHORIZATION=header)
         response = MetadataView.as_view()(request)
         response_content = json.loads(response.content)
-        self.assertListEqual(response_content['roles'], ['read', 'write', 'manage'])
+        self.assertListEqual(sorted(response_content['roles']), ['manage', 'read', 'write'])
```
kvanhijf, line 417: a Set could have accomplished the same result (Don't know whether conversion to a Set is faster than sorting)

khenderick, line 417: A set is indeed slightly faster, but it would not validate a possibly duplicate role in the return value.

PR #642, opened 2016-06-20
New unittest for set_as_template functionality

kvanhijf:
Additional set_as_template and create_from_template GUI improvements


kvanhijf:
Removed the 'approved' label because forgot to make changes to our unittest library

PR #668, opened 2016-06-24
spec-remove-hypervisor-dependency

wimpers:


wimpers:
Still working on this.

specs/remove-hypervisor-dependency.md
```diff
@@ -0,0 +1,45 @@
+# Remove hypervisor dependencies
+
+## Scope
+We want to make Open vStorage a real storage product which no longer depends on a hypervisor. In previous versions we always assumed that the Volume Driver ran together with a hypervisor (VMware/KVM). With the Edge this no longer the case as the edge client runs on the compute nodes (hypervisor) and the volume driver runs separately on a different node.
+With th FR we want to break this hypervisor dependency. This means you can now install Open vStorage and create a vdisk and add the hypervisor later.
```
khenderick, line 5: "... and add the hypervisor later". We had decided that there won't be any hypervisors left wrt our product so you can't "add" it later. Unless you mean that you can use a hypervisor of your liking, but not integrated with OVS whatsoever.

specs/remove-hypervisor-dependency.md
```diff
@@ -0,0 +1,45 @@
...
+ * Add/extend vPool: remove auto-configuration screen
+* vMachines: remove ity completely
+* vDisks
+ * Overview: remove vMachine column add Edge column
+ * Detail: update breadcrum to remove vMachine, remove vMachine
+* Remove vMachine Templates 
+* Administration remove hypervisor mgmt
+
+## API & DAL
+* remove vmachine and hypervisor mgtm center
```
khenderick: The PMachine object will also be removed.

specs/remove-hypervisor-dependency.md
wimpers, line 5: the latter, not integrated with OVS

specs/remove-hypervisor-dependency.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Hypervisors
+### NFS
+* Remove the NFS path for VMware (VMware will be supported through iSCSI)
+### KVM
+* For KVM we need to still support the FUSE filesystem approach
+* When creating a vDisk on the fuse layer we don't interact with th hypervisor to find out which VM it is attached too.
+* .raw only
+### Edge
+* The Edge should be able to work without the FUSe layer
```
saelbrec, line 18: Wondering why this requirement is here as I think we don't want to do any changes to the Edge today? If I am not mistaken the Edge control path uses same functions as filesystem (like eg fs.create_volume).

khenderick, line 18: I think @wimpers meant that the framework should try to not rely on the FUSE mountpoint as much as possible.

wimpers, line 18: Correct. We want to remove the fuse in a later stage

PR #697, opened 2016-07-05
Allow unattended RDMA

khenderick:


kvanhijf:
The `resume_config` should also be extended with this functionality

PR #715, opened 2016-07-12
#158 - GUI speed enhancements

kvanhijf:
Global ALBA backends no longer show the pie chart
Local ALBA backends no longer show individual vPool usages

### DIFF
--- a/webapps/frontend/app/views/site/dashboard.html
+++ b/webapps/frontend/app/views/site/dashboard.html
@@ -36,7 +36,7 @@ but WITHOUT ANY WARRANTY of any kind.
                 </div>
                 <div class="panel-body">
                     <div data-bind="lazylist: {
-                                        displaymode: 'table', items: vPools, itemsLoaded: 'vPoolsLoaded',
+                                        displaymode: 'table', items: vPools, itemsLoaded: vPoolsLoaded(),
                                         loadedObservable: 'loaded', colspan: 5
                                     }" >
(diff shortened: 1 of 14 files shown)


webapps/frontend/app/widgets/lazylist/view.html
```diff
@@ -28,31 +28,38 @@
...
-                            <i data-i18n="ovs:generic.noentries"></i>
-                        </td>
-                    </tr>
-                </tbody>
-            <!-- /ko -->
-            <!-- ko if: items().length > 0 -->
-                <tbody data-bind="foreach: items">
-                    <!-- ko if: !$parent.isLoaded($data) -->
+        <!-- ko if: !itemsHaveBeenLoaded($parent) -->
+            <div>
```
khenderick: Might be better to keep it as table (with e.g. the headers displayed) just like all our other tables.

webapps/frontend/app/views/site/dashboard.html
```diff
@@ -36,7 +36,7 @@
                 </div>
                 <div class="panel-body">
                     <div data-bind="lazylist: {
-                                        displaymode: 'table', items: vPools,
+                                        displaymode: 'table', items: vPools, itemsLoaded: 'vPoolsLoaded',
```
khenderick: Why not just pass in the observable itself, just like `vPools`?

PR #728, opened 2016-07-15
Make sure a vDisk that is cleaned up

khenderick:


ovs/lib/vdisk.py
```diff
@@ -81,6 +81,19 @@ def list_volumes(vpool_guid=None):
         return response
 
     @staticmethod
+    def _clean_vdisk_from_model(vdisk):
+        """
+        Removes a vDisk from the model
+        :param vdisk_guid: The vDisk to be removed
```
kvanhijf: Param name is incorrect

Issue #723, opened 2016-07-15
Disabling DTL optimized


kvanhijf:
The framework's way of disabling the DTL for a vPool is not correct
We set None as value, but due to json parsing and volumedriver parsing, this ends up as a string "null" in the volumedriver's config, resulting in the volumedriver trying to connect to this IP, but failing obviously

Question now arises for @redlicha : Why don't we get a DEGRADED dtl status when configuring the IP as "null"?

The solution is to pass an empty string for 'fs_dtl_host' instead of None

PS: Investigate disabling DTL for a single volume too, whether this is correctly done

redlicha:
The volumedriver's parsing does not play a role here: it receives a json config with the value of `fs_dtl_host` set to the string `"null"`, which it then happily interprets as the address (hostname) of the DTL.
WRT the state: it's `OK_STANDALONE`, to be investigated - https://github.com/openvstorage/volumedriver/issues/68

khenderick:
To be clear here, inside the volumedriver's configuration file, in the `filesystem` section, we do not set `fs_dtl_host` to `"null"`, but to `null`, so it's not set to a string. It must thus be the volumedriver's json parser which cannot handle null values and somehow casts this as a string.

```
root@ovs100:~# etcdctl get /ovs/vpools/a904f2c6-d064-42b1-9688-8de6e0a2ae3a/hosts/nodtlsf2XSrGzR4aVijtI/config/filesystem
{
    "fs_raw_disk_suffix": ".raw", 
    "fs_enable_shm_interface": 0, 
    "fs_metadata_backend_arakoon_cluster_nodes": [], 
    "fs_metadata_backend_mds_nodes": [
        {
            "host": "10.100.169.100", 
            "port": 26301
        }
    ], 
    "fs_metadata_backend_type": "MDS", 
    "fs_virtual_disk_format": "raw", 
    "fs_enable_network_interface": 1, 
    "fs_dtl_config_mode": "Manual", 
    "fs_dtl_host": null
}
```

In the volumedriver's logs:

```
...
    },
    "filesystem": {
        "fs_raw_disk_suffix": ".raw",
        "fs_enable_shm_interface": "0",
        "fs_metadata_backend_arakoon_cluster_nodes": "",
        "fs_metadata_backend_mds_nodes": [
            {
                "host": "10.100.169.100",
                "port": "26301"
            }
        ],
        "fs_metadata_backend_type": "MDS",
        "fs_virtual_disk_format": "raw",
[...]

PR #751, opened 2016-07-29
Make sure we don't use the string "null" instead of null, None or "".

khenderick:


khenderick:
First implementation of #723, but more to be discussed and possibly implemented.

kvanhijf:
Fixed by https://github.com/openvstorage/framework/issues/723

PR #752, opened 2016-07-29
Block roles to be assigned to md device members

khenderick:


kvanhijf:
Shouldn't this check also be implemented in the StorageRouterController?

Issue #760, opened 2016-08-02
Eugene-updates celery ping issue with FQDN


jeroenmaelbrancke:
When the celery process is register with the FQDN the heartbeat doesn't work anymore.

example:

```
root@cpu-01:~# /usr/bin/celery status -b amqp://ovs:0penv5tor4ge@10.107.1.22
celery@stor-04.be-g8-3: OK
celery@cpu-04.be-g8-3: OK
celery@cpu-01.be-g8-3: OK
celery@cpu-03.be-g8-3: OK
celery@cpu-02.be-g8-3: OK
celery@stor-02.be-g8-3: OK
celery@stor-03.be-g8-3: OK
celery@stor-01.be-g8-3: OK
```

The if 'celery@{0}: OK'.format(node.name) in worker_states is not working.
Using a regex would solve this issue.

/opt/OpenvStorage/ovs/extensions/generic/heartbeat.py:

```
celery_path = OSManager.get_path('celery')
worker_states = check_output("{0} inspect ping -b {1} --timeout=5 2> /dev/null | grep OK | perl -pe 's/\x1b\[[0-9;]*m//g' || true".format(celery_path, amqp), shell=True)
routers = StorageRouterList.get_storagerouters()
for node in routers:
    if node.heartbeats is None:
        node.heartbeats = {}
    if 'celery@{0}: OK'.format(node.name) in worker_states:
        node.heartbeats['celery'] = current_time
```

wimpers:
@kvanhijf did we package this already? If packaged please close

kvanhijf:
Not yet packaged because there are currently permission issues when trying to package
Will need @khenderick for this

kvanhijf:
Packaged by @khenderick after having fixed the packaging problem

pploegaert:
validated by @jeroenmaelbrancke on gig setup

PR #761, opened 2016-08-02
Used a regex for celery heartbeat.

jeroenmaelbrancke:
Issue #760
@kvanhijf can you do a code review?

### DIFF
--- a/ovs/extensions/generic/heartbeat.py
+++ b/ovs/extensions/generic/heartbeat.py
@@ -33,25 +33,23 @@ amqp = '{0}://{1}:{2}@{3}//'.format(Configuration.get('ovs.core.broker.protocol'
 
 celery_path = OSManager.get_path('celery')
 worker_states = check_output("{0} inspect ping -b {1} --timeout=5 2> /dev/null | grep OK | perl -pe 's/\x1b\[[0-9;]*m//g' || true".format(celery_path, amqp), shell=True)
-domain = check_output(["hostname", "-d"]).splitlines()[0]
+domain = check_output("hostname -d", shell=True).strip()
 routers = StorageRouterList.get_storagerouters()
 for node in routers:
     if node.heartbeats is None:
         node.heartbeats = {}
-    for s in worker_states.splitlines():
-        found = False
-        if re.match("^-> celery@{0}:".format(node.name), s):
-            if re.match("^-> celery@{0}: OK$".format(node.name), s):
-                found = True
-        if found:
-            node.heartbeats['celery'] = current_time
-            continue
+    for worker_state in worker_states.splitlines():
+        if re.match("^-> celery@{0}:".format(node.name), worker_state):
+            if re.match("^-> celery@{0}: OK$".format(node.name), worker_state):
+                node.heartbeats['celery'] = current_time
+                break
+            break
         else:
             if len(domain) == 0:
                 continue
-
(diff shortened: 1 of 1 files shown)


ovs/extensions/generic/heartbeat.py
```diff
@@ -32,12 +33,26 @@
 
 celery_path = OSManager.get_path('celery')
 worker_states = check_output("{0} inspect ping -b {1} --timeout=5 2> /dev/null | grep OK | perl -pe 's/\x1b\[[0-9;]*m//g' || true".format(celery_path, amqp), shell=True)
+domain = check_output(["hostname", "-d"]).splitlines()[0]
```
kvanhijf: do a `strip()` here instead of `splitlines()[0]` and always use single quotes please in Python
Alternatively you can use `check_output('hostname -d', shell=True).strip()`

ovs/extensions/generic/heartbeat.py
```diff
@@ -32,12 +33,26 @@
...
 routers = StorageRouterList.get_storagerouters()
 for node in routers:
     if node.heartbeats is None:
         node.heartbeats = {}
-    if 'celery@{0}: OK'.format(node.name) in worker_states:
-        node.heartbeats['celery'] = current_time
+    for s in worker_states.splitlines():
```
kvanhijf: Never use single character variables please, make sure the variables have a meaningful name

ovs/extensions/generic/heartbeat.py
```diff
@@ -32,12 +33,26 @@
...
+        found = False
+        if re.match("^-> celery@{0}:".format(node.name), s):
+            if re.match("^-> celery@{0}: OK$".format(node.name), s):
```
kvanhijf: `found` should be set to True as soon as its found, regardless of the status being OK
If the status is not OK, the celery heartbeat should not be updated, but we shouldn't attempt to check again without fqdn

ovs/extensions/generic/heartbeat.py
```diff
@@ -32,12 +33,26 @@
...
+                found = True
+        if found:
+            node.heartbeats['celery'] = current_time
+            continue
+        else:
```
kvanhijf: Obsolete `else`stack since its right after a `continue`

ovs/extensions/generic/heartbeat.py
```diff
@@ -32,12 +33,26 @@
...
+            if len(domain) == 0:
+                continue
+
+            if re.match("^-> celery@{0}.{1}:".format(node.name, domain), s):
```
kvanhijf: Also you have to make sure that celery returns OK, otherwise the heartbeat should not be updated

ovs/extensions/generic/heartbeat.py
```diff
@@ -32,12 +33,24 @@
...
         node.heartbeats = {}
-    if 'celery@{0}: OK'.format(node.name) in worker_states:
-        node.heartbeats['celery'] = current_time
+    for worker_state in worker_states.splitlines():
+        if re.match("^-> celery@{0}:".format(node.name), worker_state):
+            if re.match("^-> celery@{0}: OK$".format(node.name), worker_state):
+                node.heartbeats['celery'] = current_time
+                break
+            break
+        else:
```
kvanhijf: Can be shortened to
`elif domain != '':`

PR #781, opened 2016-08-06
FWK-763 - reduced etcd values + added limitation to snapshots count

pploegaert:


pploegaert:
openvstorage-2.7.2-rev.3835.9f486e5

PR #807, opened 2016-08-12
Process 401 errors correctly

khenderick:


ovs/extensions/api/client.py
```diff
@@ -129,23 +129,43 @@ def _process(self, response, overrule_raw=False):
...
+        parsed_output = None
         try:
-            return_data = response.json()
-            return return_data
+            parsed_output = response.json()
         except:
-            raise RuntimeError('Could not parse returned data: {0}: {1}'.format(response.status_code, response.text))
+            pass
+
+        if 200 <= response.status_code < 300:
```
kvanhijf: Use `status_code` here instead

ovs/extensions/api/client.py
```diff
@@ -129,23 +129,43 @@ def _process(self, response, overrule_raw=False):
...
         except:
-            raise RuntimeError('Could not parse returned data: {0}: {1}'.format(response.status_code, response.text))
+            pass
+
+        if 200 <= status_code < 300:
+            if parsed_output is not None:
+                return parsed_output
+            raise RuntimeError('Could not parse returned data: {0}: {1}'.format(status_code, response.text))
+        else:
+            if response.status_code in [401, 403]:
```
domsj: Use `status_code` here instead

khenderick:
Apparently, @domsj is a better reviewer than @kvanhijf.

Issue #824, opened 2016-08-19
User Mgmt: add a better indication that the OAuth 2 Clients are for a user group


JeffreyDevloo:
## Problem description

The only indication that the OAuth 2 Client is for a specific user group is the blue bar on the top of the screen.
![selection_023](https://cloud.githubusercontent.com/assets/17570109/17808657/9113ff88-6611-11e6-8fc0-ecfbfb9a690b.png)
## Proposed solution

Add a small indicator in the title aswell.
![selection_024](https://cloud.githubusercontent.com/assets/17570109/17808710/de74005c-6611-11e6-81f5-6f9608bbdcbb.png)
### Setup

Hyperconverged setup
- Three nodes with each three disks for the back-end
### Package information
- openvstorage                       2.7.2-rev.3859.42b3488-1   amd64        openvStorage
- openvstorage-backend               1.7.2-rev.675.37ca5b8-1    amd64        openvStorage Backend plugin
- openvstorage-backend-core          1.7.2-rev.675.37ca5b8-1    amd64        openvStorage Backend plugin core
- openvstorage-backend-webapps       1.7.2-rev.675.37ca5b8-1    amd64        openvStorage Backend plugin Web Applications
- openvstorage-cinder-plugin         1.2.2-rev.32.948a8c1-1 amd64        OpenvStorage Cinder plugin for OpenStack
- openvstorage-core                  2.7.2-rev.3859.42b3488-1   amd64        openvStorage core 
- openvstorage-hc                    1.7.2-rev.675.37ca5b8-1    amd64        openvStorage Backend plugin HyperConverged
- openvstorage-health-check          2.0.0-rev.117.3212ea9-1    amd64        Open vStorage HealthCheck
- openvstorage-sdm                   1.6.2-rev.330.f06c8de-1    amd64        Open vStorage Backend ASD Manager
- openvstorage-test                  2.7.2-rev.980.0dbc80f-1    amd64        openvStorage autotest suite
- openvstorage-webapps               2.7.2-rev.3859.42b3488-1   amd64        openvStorage Web Applications

wimpers:
@JeffreyDevloo  is this something you can pick up?

JeffreyDevloo:
Fixed in openvstorage-2.7.2-rev.3880.93db3d4

PR #840, opened 2016-08-23
Changed the title in users.html to fit the better indication need

JeffreyDevloo:
## https://github.com/openvstorage/framework/issues/824

Adds the following indication to the page:
selection_049


webapps/frontend/app/views/site/users.html
```diff
@@ -95,7 +95,14 @@ <h2 class="pull-left" data-i18n="ovs:users.users.title"></h2>
         </table>
     </div>
     <div>
-        <h2 class="pull-left" data-i18n="ovs:users.clients.title"></h2>
+        <h2 class="pull-left">
+            <span data-i18n="ovs:users.clients.title"></span>
+            <!-- ko if: selectedUser() -->
+            <span>for</span>
```
khenderick: This should be indented, and `for` should be translated.

PR #859, opened 2016-08-29
Domain and recovery domain need an MDS

wimpers:
…ge/framework/issues/791

### DIFF
--- a/ovs/lib/tests/vdisk_tests/test_get_set_config_params.py
+++ b/ovs/lib/tests/vdisk_tests/test_get_set_config_params.py
@@ -49,7 +49,7 @@ class VDiskTest(unittest.TestCase):
         fakesleep.monkey_patch()
         EtcdConfiguration.set('/ovs/framework/storagedriver|mds_tlogs', 100)
         EtcdConfiguration.set('/ovs/framework/storagedriver|mds_maxload', 75)
-        EtcdConfiguration.set('/ovs/framework/storagedriver|mds_safety', 2)
+        EtcdConfiguration.set('/ovs/framework/storagedriver|mds_safety', 3)
 
     def setUp(self):
         """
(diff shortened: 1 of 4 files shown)


ovs/lib/tests/vdisk_tests/test_get_set_config_params.py
```diff
@@ -49,7 +49,7 @@ def setUpClass(cls):
         fakesleep.monkey_patch()
         EtcdConfiguration.set('/ovs/framework/storagedriver|mds_tlogs', 100)
         EtcdConfiguration.set('/ovs/framework/storagedriver|mds_maxload', 75)
-        EtcdConfiguration.set('/ovs/framework/storagedriver|mds_safety', 2)
+        EtcdConfiguration.set('/ovs/framework/storagedriver|mds_safety', 3)
```
khenderick: Does this change the behavior of the tests? Are you sure the test still validates what is expected? Noticed any coverage changes?

wimpers: Valid point. Test will indeed have to be reviewed. The creation of a lot of disks will fail already as the mds service isn't available on 3 nodes. Maybe better to also extend the test to check if mds safety is as the one set is actually achieved.

wimpers: Would it be wise to update the config value in the actual code to 3 and leave the tests with 2 as default?

khenderick: It's a unittest, there are no volumes or mds services actually running, the test just validates the behavior of the framework's decisions. It might indeed be a good idea to revert these two test changes, as there are separate tests that validate the mds_safety behavior.

wimpers: We should also add a check in the tests to make sure it validates the distribution of the mds across domain and recovery domains.

khenderick: That's already in there: https://github.com/openvstorage/framework/blob/master/ovs/lib/tests/mdsservice_tests/test_mdsservice.py

Issue #777, opened 2016-08-05
Noticed list index out of range on ovs.vdisk.dtl_checkup in ovs_workers.log


pploegaert:
2016-08-05 11:36:56 35800 +0200 - ovs52n252 - 125179/140635645032256 - celery/celery.worker.job - 977 - ERROR - Task ovs.vdisk.dtl_checkup[61e2e1d9-cc37-40a7-b68d-0361671f41d7] raised unexpected: IndexError('list index out of range',)
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(_args, *_kwargs)
  File "/usr/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in **protected_call**
    return self.run(_args, *_kwargs)
  File "/opt/OpenvStorage/ovs/lib/helpers/decorators.py", line 236, in new_function
    output = function(_args, *_kwargs)
  File "/opt/OpenvStorage/ovs/lib/vdisk.py", line 956, in dtl_checkup
    current_sr = [sd for sd in vpool.storagedrivers if sd.storage_ip == current_dtl_config.host][0].storagerouter
IndexError: list index out of range

matthiasdeblock:
Also, lot of these messages on GIG's env. (unstable)

```
2016-09-07 14:59:58 95300 +0200 - ovs-04.be-gen8-2 - 37168/140206635284288 - celery/celery.worker.job - 429 - ERROR - Task ovs.vdisk.dtl_checkup[faff60a2-27bb-438f-abc8-a0a501eac429] raised unexpected: IndexError('list index out of range',)
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/OpenvStorage/ovs/lib/helpers/decorators.py", line 236, in new_function
    output = function(*args, **kwargs)
  File "/opt/OpenvStorage/ovs/lib/vdisk.py", line 911, in dtl_checkup
    current_sr = [sd for sd in vpool.storagedrivers if sd.storage_ip == current_dtl_config.host][0].storagerouter
IndexError: list index out of range
```

matthiasdeblock:
[...]

PR #863, opened 2016-08-31
FWK-777 - fix list index out of range

pploegaert:


JeffreyDevloo:
Edit: No longer approved. After re-evaluating the code, I made a mistake on what the issue could be.
Need to test this fix further before approving.

khenderick:
Just wondering; did you guys test the rest of the changed code? Is the `dtl_checkup` still doing what it's supposed to do when this list would end up empty?

ovs/lib/vdisk.py
```diff
@@ -908,7 +908,9 @@ def dtl_checkup(vpool_guid=None, vdisk_guid=None, storagerouters_to_exclude=None
                 current_sr = None
                 available_primary_srs = available_primary_srs.difference(available_secondary_srs)
                 if current_dtl_config is not None:
-                    current_sr = [sd for sd in vpool.storagedrivers if sd.storage_ip == current_dtl_config.host][0].storagerouter
+                    current_srs = [sd for sd in vpool.storagedrivers if sd.storage_ip == current_dtl_config.host]
```
khenderick, line 911: `current_srs` should be `current_sds`, as we're working with StorageDrivers here, not StorageRouters.

ovs/lib/vdisk.py
```diff
@@ -908,7 +908,9 @@ def dtl_checkup(vpool_guid=None, vdisk_guid=None, storagerouters_to_exclude=None
...
+                    if current_srs:
```
JeffreyDevloo, line 912: I'd opt for len(current_srs)>=1 for more clarification what its about

khenderick:
Closed since this PR only fixes the error itself, and not the reason why it was caused. Will be resolved by another PR implementing the discussed requirements in #777

PR #865, opened 2016-08-31
FWK- 762 configure_disk failed IndexError

wimpers:


khenderick:
Why got this approved? This change doesn't fix the reported error:

```
...
    first_element = value['values'][0]['timestamp']
IndexError: list index out of range
```

Since we're getting an `IndexError` and not a `KeyError`, we know that `values` is in `value`, and the `IndexError` is on the `[0]`, so the "fix" doesn't do anything.

khenderick:
Apart from that, just "hiding" the error in this case might make things worse. The fact that there's nothing in that list is unexpected and something must go wrong elsewhere. If we just ignore this error and move on, it might be possible that tasks get executed at the same time (=issues) or not at all (=issue).

PR #879, opened 2016-09-06
Use the ObjectRegistry to load a vDisk's object type

khenderick:


ovs/dal/hybrids/vdisk.py
```diff
@@ -296,3 +300,15 @@ def calculate_delta(key, dynamic, current_stats):
...
+
+    def reload_clients(self, client):
+        """
+        Reloads the StorageDriverClient and ObjectRegistryClient
+        """
+        if self.vpool_guid:
+            self._frozen = False
+            if client == 'storagedriver':
+                self._storagedriver_client = StorageDriverClient.load(self.vpool)
+            if client == 'objectregistry':
```
kvanhijf: `elif` here is more optimal

ovs/dal/hybrids/vdisk.py
```diff
@@ -296,3 +300,15 @@ def calculate_delta(key, dynamic, current_stats):
                 else:
                     current_stats['{0}_ps'.format(key)] = max(0, (current_stats[key] - previous_stats[key]) / delta)
         volatile.set(prev_key, current_stats, dynamic.timeout * 10)
+
+    def reload_clients(self, client):
+        """
+        Reloads the StorageDriverClient and ObjectRegistryClient
+        """
+        if self.vpool_guid:
+            self._frozen = False
```
kvanhijf, line 311: I don't understand this `self._frozen` logic here. If it acts like a lock, shouldn't it be the other way around then?
`frozen = True` before re-establishing the client and `frozen = False` afterwards

khenderick: The `self._frozen` is used internal to prevent (some) properties from being set. Not (temporarily) setting it to `False` would cause an error when the clients are changed. It's a I-know-what-I'm-doing-here flag.

khenderick:
Still have to fix the unittests here

ovs/dal/hybrids/vdisk.py
```diff
@@ -296,3 +302,15 @@ def calculate_delta(key, dynamic, current_stats):
                 else:
                     current_stats['{0}_ps'.format(key)] = max(0, (current_stats[key] - previous_stats[key]) / delta)
         volatile.set(prev_key, current_stats, dynamic.timeout * 10)
+
+    def reload_clients(self, client):
```
kvanhijf, line 306: Function name does not cover the load
A rename was unnecessary imo, since you only reload 1 type of client, so `reload_client` was just fine

khenderick: Agree, `reload_client` might have been more appropriate, but at least, it's consistent with `vPool`. Feel free to rename in a future change, but I won't make a new PR just for this renaming.

PR #886, opened 2016-09-08
Increase volume limit to 64TiB

khenderick:


ovs/lib/vdisk.py
```diff
@@ -1153,7 +1153,10 @@ def _set_vdisk_metadata_pagecache_size(vdisk):
         storagedriver_config.load()
         metadata_page_capacity = 256
         cluster_size = storagedriver_config.configuration.get('volume_manager', {}).get('default_cluster_size', 4096)
-        num_pages = int(vdisk.size / (metadata_page_capacity * cluster_size))
+        num_pages = min(
```
kvanhijf: This can be optimized / simplified

kvanhijf (changes requested):

kvanhijf (approved):

Issue #762, opened 2016-08-02
configure_disk failed IndexError: list index out of range


JeffreyDevloo:
We saw this on a hyperconverged node during role configuration:

```
2016-08-02 13:27:08 15500 +0200 - cmp03 - 12223/140461375452992 - celery/celery.worker.job - 665 - ERROR - Task ovs.storagerouter.configure_disk[84aff99f-3533-4ce2-a663-5e80dce8fe86] raised unexpected: IndexError('list index out o
f range',)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/OpenvStorage/ovs/lib/helpers/decorators.py", line 291, in new_function
    first_element = value['values'][0]['timestamp']
IndexError: list index out of range
2016-08-02 13:27:08 53800 +0200 - cmp03 - 12223/140461375452992 - celery/celery.worker.job - 666 - ERROR - Task ovs.storagerouter.configure_disk[4f0106f9-2af5-4e61-8513-412c83eda095] raised unexpected: IndexError('list index out o
f range',)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/OpenvStorage/ovs/lib/helpers/decorators.py", line 291, in new_function
    first_element = value['values'][0]['timestamp']
IndexError: list index out of range
2016-08-02 13:27:08 54600 +0200 - cmp03 - 12223/140461375452992 - celery/celery.worker.autoscale - 667 - INFO - Scaling down 1 processes.
2016-08-02 13:27:08 72900 +0200 - cmp03 - 12223/140461375452992 - celery/celery.worker.job - 668 - ERROR - Task ovs.storagerouter.configure_disk[64a132bd-eef4-4cc9-86e8-b721c91959d8] raised unexpected: IndexError('list index out o
f range',)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
[...]

PR #902, opened 2016-09-14
#762 - Index out of range when configuring disk

kvanhijf:


ovs/lib/helpers/decorators.py
```diff
@@ -285,7 +288,7 @@ def update_value(key, append, value_to_update=None):
                 # Poll the arakoon to see whether this call is the first in list, if so --> execute, else wait
                 first_element = None
                 counter = 0
-                while first_element != now and counter < timeout:
+                while counter < timeout:
                     if persistent_client.exists(persistent_key):
                         value = persistent_client.get(persistent_key)
                         first_element = value['values'][0]['timestamp']
```
khenderick: Let's catch the index error here (which can be caused by the race condition we discussed) and make sure it eventually raises the `EnsureSingleTimeoutReached` exception, which is more correct. To be determined if we need a retry mechanism in the future or not.

ovs/lib/storagerouter.py
```diff
@@ -1861,6 +1861,11 @@ def configure_disk(storagerouter_guid, disk_guid, partition_guid, offset, size,
                 if partition.mountpoint != mountpoint:
                     raise RuntimeError('Unexpected mountpoint')
             StorageRouterController._logger.debug('Mountpoint configured')
+        elif len(roles) == 0:  # Unmount disk without any roles assigned to be claimed by a backend
+            with remote(storagerouter.ip, [DiskTools], username='root') as rem:
+                StorageRouterController._logger.debug('Unmounting mountpoint: {0}'.format(partition.mountpoint))
+                rem.DiskTools.umount(mountpoint=partition.mountpoint)
```
khenderick: I'd rather not unmount the disk, we're not sure the customer doesn't use it for something else. Let's create a separate ticket where a customer can optionally check a box in the GUI where he indicates we can/should wipe the disk and - in case of role removal - unmount it.

khenderick (changes requested):

ovs/lib/storagerouter.py
kvanhijf: Requested issue: https://github.com/openvstorage/framework/issues/905

khenderick (approved):

PR #917, opened 2016-09-15
Apparently lstrip didn't work as expected

khenderick:


kvanhijf (changes requested):
Please create function for this in some helperclass or generic class. Less code duplication ftw

kvanhijf (approved):

Issue #896, opened 2016-09-13
Rollback to a snapshot


FastGeert:
Rollback to a snapshot is currently not available in Fargo. 

The proposed method is to create a clone from the snapshot and continue using this. 

Technically this is a perfect solution, but for the client of the OVS API it creates a lot of complexity as the client needs to continue with another disk with a different id.
Deleting a disk becomes more complex because maybe the client needs to delete an ancestor disk as well, ....
I don't think that removing the rollback from the OVS api is a good choice.
If rolling back to a snapshot is easier to implement and maintain, you cloud implement rollback like that internally, but keep the easier api for the client, and do the reference counting behind the scene.

Currently not blocking for us.

wimpers:
To be discussed how to tackle.

khenderick:
A few issues here:
- GUI messages related to snapshot rollback are not translated (probably moved in the translation file but not correctly reflected in code)
- The GUI passes the snapshot-to-revert's timestamp as integer, but internally it's stored as string. And that obviously doesn't match.

Shouldn't be more than 15 minutes of work to fix the above.

@FastGeert, it's not "unavailable", it just doesn't work (meaning "bug")

khenderick:
Fixed by #934, packaged in openvstorage-2.7.3-rev.4016.ea03de8

JeffreyDevloo:
### Information
We use the rollback API call in our automated tests and have not faced any issues with the API call. Then again we are a little bit behind in verification...

PR #934, opened 2016-09-20
A couple of small fixes to get snapshot rollback working

FastGeert:
After some digging in the framework source code I discovered that getting snapshot rollback working is only a couple of small changes.

This will spare us a lot of useless temporary engineering work until ovs engineers can provide a fix for rollback that is geo-scale compatible. (see https://github.com/openvstorage/framework/issues/896)


khenderick (approved):
Looks good to me, thanks!

khenderick (changes requested):
On hold due to missing CLA.

khenderick (approved):

PR #984, opened 2016-10-03
Backwards compatible change - starts all arakoons in config file

JeffreyDevloo:
Will start all arakoons in the configuration file.
Usefull in more efficient extending flow ( create -> extend -> link -> start all)


khenderick (approved):

khenderick:
Packaged in openvstorage-2.7.3-rev.4065.f286ee6

PR #1006, opened 2016-10-07
Add support for unique constraints + 'contains' statement in queries

khenderick:


ovs/dal/dataobject.py
```diff
@@ -22,10 +22,11 @@
 import time
 import json
 import inspect
+import hashlib
 from random import randint
 from ovs.dal.exceptions import (ObjectNotFoundException, ConcurrencyException, LinkedObjectException,
                                 MissingMandatoryFieldsException, RaceConditionException, InvalidRelationException,
-                                VolatileObjectException)
+                                VolatileObjectException, UniqueContraintViolationException)
```
kvanhijf: whats a contraint?

kvanhijf (changes requested):

khenderick:
After some discussions with @kvanhijf, I'll add following:
- A hybrid test to ensure that only supported fields are configured to be unique
- Go through all our objects to add more constraints if required
- Add migration code (if not, an upgraded unique field cannot be changed anymore)

kvanhijf (approved):
Appears kosjer

PR #1033, opened 2016-10-14
A bunch of OVSClient fixes/changes

khenderick:
Support @return_plain + Add delete call + Suppress warnings


dejonghb (approved):
Fine for me

dejonghb (approved):

PR #1076, opened 2016-10-21
Fixed incorrect handling of vDisk create and clone failures

khenderick:


ovs/lib/vdisk.py
```diff
@@ -530,17 +522,26 @@ def create_from_template(vdisk_guid, name, storagerouter_guid=None):
...
             MDSServiceController.ensure_safety(new_vdisk)
             VDiskController.dtl_checkup.delay(vdisk_guid=new_vdisk.guid)
             VDiskController._set_vdisk_metadata_pagecache_size(new_vdisk)
         except Exception as ex:
             VDiskController._logger.error('Clone vDisk on volumedriver level failed with exception: {0}'.format(str(ex)))
-            try:
-                VDiskController.delete(new_vdisk.guid)
-            except Exception as ex2:
-                VDiskController._logger.exception('Exception during exception handling of "create_clone_from_template" : {0}'.format(str(ex2)))
+            if volume_id is not None:
```
kvanhijf: Its still necessary to clean up model in case `ensure_safety` for example fails

kvanhijf (changes requested):

khenderick:
Need some more thought

Issue #972, opened 2016-09-30
Set_config_params incorrectly assumes a DTL location should be provided


kvanhijf:
When the storagerouters of an environment have been assigned to domains and a vdisk volumedrivers details are changed (without specifying a DTL location), the call will fail
ValueError: If DTL mode is Asynchronous or Synchronous, a Domain guid should always be specified

Afaik when 'Nothing Selected' is visible in the GUI for a vDisk, this implies that the DTL location can be chosen by the framework, however, the code reacts differently and raises above errormessage

So we should be able to handle passing on no locations and perhaps add a message in the GUI mentioning that 'Nothing selected' means 'I don't care'

wimpers:
Might be good to fix https://github.com/openvstorage/framework/issues/849 at the same go

kvanhijf:
https://github.com/openvstorage/framework/pull/1082 --> openvstorage-2.7.4-rev.4195.998359c

kvanhijf:
The ticket linked by @wimpers (https://github.com/openvstorage/framework/issues/849) has not been fixed by this fix

wimpers:
Closed by PM.

PR #1082, opened 2016-10-25
#972 - Setting config params for vDisk no longer raises when no domai…

kvanhijf:
…ns specified


khenderick (approved):

kvanhijf:
openvstorage-2.7.4-rev.4195.998359c

PR #1106, opened 2016-10-28
Allow to skip object caching when the cache lock can not be acquired

khenderick:


ovs/dal/tests/test_basic.py
```diff
@@ -1522,3 +1522,18 @@ def test_dynamic_unicode_return(self):
         disk.dynamic_int = 5
         dynamic = disk.updatable_int
         self.assertIsInstance(dynamic, int, 'Return value should be int')
+
+    def test_aquired_lock_during_caching(self):
```
kvanhijf: acquired*

ovs/dal/tests/test_basic.py
```diff
@@ -1522,3 +1522,18 @@ def test_dynamic_unicode_return(self):
...
+        """
+        Validates whether loading an object won't fail on a non-aquireable cache lock
```
kvanhijf: Acquireable*

kvanhijf (changes requested):

kvanhijf (approved):

PR #1107, opened 2016-10-31
#134 - Enhancements to the `monitor services` output for both ASD-man…

kvanhijf:
…ager and OVS

### DIFF
--- a/ovs/lib/setup.py
+++ b/ovs/lib/setup.py
@@ -24,6 +24,8 @@ import sys
 import json
 import time
 import signal
+from subprocess import CalledProcessError, check_output
+from paramiko import AuthenticationException
 from ovs.dal.hybrids.servicetype import ServiceType
 from ovs.extensions.db.arakoon.ArakoonInstaller import ArakoonClusterConfig, ArakoonInstaller
 from ovs.extensions.db.arakoon.configuration import ArakoonConfiguration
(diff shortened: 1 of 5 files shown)


ovs/lib/setup.py
```diff
@@ -779,6 +780,68 @@ def remove_node(node_ip, silent=None):
...
 
     @staticmethod
+    def monitor_services():
+        """
+        Monitor the OVS services
+        :return: None
+        """
+        try:
+            previous_output = None
+            service_manager = check_output('cat /proc/1/comm', shell=True).strip()
```
khenderick: Why not offload this detection to the ServiceManager, then we have this detection code in only one place. Apart from that, I think it might be a better idea to move this code to the StorageRouterController as it "monitors" services running on a StorageRouter, and has not much to do with the Setup.

khenderick: Or maybe put this code completely in the ServiceManager?

khenderick (approved):

kvanhijf:
https://github.com/openvstorage/framework/pull/1107 --> openvstorage-2.7.4-rev.4241.ac52433
https://github.com/openvstorage/alba-asdmanager/pull/143 --> openvstorage-sdm-1.6.4-rev.412.7a2713e

Issue #854, opened 2016-08-26
vDisk gui work


wimpers:
On a vDisk detail page remove the caching section (strategy, dedupe, read cache limit)

kvanhijf:
https://github.com/openvstorage/framework/pull/1109 --> openvstorage-2.7.4-rev.4243.1b1eb26

kvanhijf:
@QA: VDiskController.get_config_params should no longer return any information regarding caching: Caching strategy, caching behavior, readcache limit

In the GUI, on the management actions tab, all 'caching' information should be gone

JeffreyDevloo:
### Steps
- Setup latest cluster
- Added disk roles
- Added backend
- Initialized disks
- Claimed ASDs (default preset was satisfied)
- Added vpool
- Added vdisk
- Checked the vdisk pages and API

### API test code

Test the fetching of the information (as mentioned in the ticket) and the changing of the caching (editing the values of the cache - test for backwards compatibility eg. old api still functions)(Not mentioned in the ticket)

```
# Copyright (C) 2016 iNuron NV
#
# This file is part of Open vStorage Open Source Edition (OSE),
# as available from
#
#      http://www.openvstorage.org and
#      http://www.openvstorage.com.
#
# This file is free software; you can redistribute it and/or modify it
# under the terms of the GNU Affero General Public License v3 (GNU AGPLv3)
# as published by the Free Software Foundation, in version 3 as it comes
# in the LICENSE.txt file of the Open vStorage OSE distribution.
#
# Open vStorage is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY of any kind.

from ci.helpers.api import OVSClient

class VDiskTest(object):

    def __init__(self):
        self.api = OVSClient('10.100.199.151', 'admin', 'admin')

    def run(self):
        # Run all test cases
        self.test_get_config_params()

        self.test_set_config_params(old=False)
        self.test_set_config_params(old=True)

    def _get_config_params(self, vdisk_guid):
        task_guid = self.api.get('/vdisks/{0}/get_config_params'.format(vdisk_guid))

        task_result = self.api.wait_for_task(task_id=task_guid, timeout=30)
        if not task_result[0]:
            error_msg = "Fetching config_params for vdisk '{0}' failed with '{1}'".format(vdisk_guid, task_result[1])
            raise RuntimeError(error_msg)
        return task_result[1]

    def _set_config_params(self, vdisk_guid, data):
[...]

PR #1113, opened 2016-11-02
#854 - Made set_config_params API backwards compatible

kvanhijf:


webapps/api/backend/views/vdisks.py
```diff
@@ -118,6 +118,10 @@ def set_config_params(self, vdisk, new_config_params, version):
                 raise HttpNotAcceptableException(error_description='API version 1 requires a Storage Router IP',
                                                  error='invalid_version')
             new_config_params['dtl_target'] = [junction.domain_guid for junction in storage_router.domains]
+        if version >= 6:
```
khenderick: Why would we need this check? We just check if the values are in there and remove them, even with newer clients (where they are not in there anyway). And since the VDiskController is always the latest code, it should always be removed.

khenderick (changes requested):

khenderick (approved):

kvanhijf:
openvstorage-2.7.4-rev.4254.b181bf9

PR #1125, opened 2016-11-04
Correctly pass through raw version information when relaying calls

khenderick:


webapps/api/backend/decorators.py
```diff
@@ -150,8 +150,10 @@ def new_function(*args, **kwargs):
             version_match = regex.match(request.META['HTTP_ACCEPT'])
             if version_match is not None:
                 version = version_match.groupdict()['version']
+                raw_version = version
             else:
                 version = settings.VERSION[-1]
+                raw_version = version
```
kvanhijf: `if ...:  raw_version = version
else: raw_version = version`
Simplification ftw

kvanhijf (changes requested):

kvanhijf (approved):

PR #1137, opened 2016-11-08
More secure shell

khenderick:


ovs/extensions/generic/system.py
```diff
@@ -163,7 +163,7 @@ def get_free_ports(selected_range, exclude=None, nr=1, client=None):
         if client is None:
             output = check_output(cmd, shell=True)
         else:
-            output = client.run(cmd)
+            output = client.run(cmd.split(' '))
```
kvanhijf: Its better to use `cmd.split()`, because `split()` will strip all whitespaces and `split(' ')` will keep whitespaces
Eg:

```
In [3]: cmd = 'cat  /proc/sys/net/ipv4/ip_local_port_range'   <-- 2 spaces

In [4]: cmd.split(' ')
Out[4]: ['cat', '', '/proc/sys/net/ipv4/ip_local_port_range']
```

ovs/extensions/os/centos.py
```diff
@@ -40,7 +40,7 @@ def get_path(binary_name):
         config_location = '/ovs/framework/hosts/{0}/paths|{1}'.format(machine_id, binary_name)
         if not Configuration.exists(config_location):
             try:
-                path = check_output('which {0}'.format(binary_name), shell=True).strip()
+                path = check_output('which {0}'.format(binary_name.replace(r"'", r"'\''")), shell=True).strip()
```
kvanhijf: Should be

```
"which '{0}'" ....
```

ovs/extensions/os/ubuntu.py
```diff
@@ -40,7 +40,7 @@ def get_path(binary_name):
         config_location = '/ovs/framework/hosts/{0}/paths|{1}'.format(machine_id, binary_name)
         if not Configuration.exists(config_location):
             try:
-                path = check_output('which {0}'.format(binary_name), shell=True).strip()
+                path = check_output('which {0}'.format(binary_name.replace(r"'", r"'\''")), shell=True).strip()
```
kvanhijf: Should be

```
"which '{0}'" ....
```

kvanhijf (changes requested):

kvanhijf (approved):

PR #1168, opened 2016-11-17
Write files consistent

khenderick:


ovs/extensions/generic/sshclient.py
```diff
@@ -523,20 +523,27 @@ def file_write(self, filename, contents, mode='w'):
         :param contents: Contents to write to the file
         :param mode: Mode to write to the file, can be a, a+, w, w+
         """
+        temp_filename = '{0}~'.format(filename)
         if self.is_local is True:
-            with open(filename, mode) as the_file:
+            with open(temp_filename, mode) as the_file:
```
domsj: what happens if a bigger temp_file_name file still exists when this operation is started?

khenderick: By default (unless the `mode` is overwritten) the file will be truncated if it already exists.

khenderick: As usual, I used `strace` to verify whether everything was executed as expected.

toolslive: so the caller can provide bad input for mode ?
What happens if mode `"a+"` is used ?

khenderick: True, other modes don't make a lot of sense. Removed the option to override it.

ovs/extensions/generic/system.py
```diff
@@ -173,54 +173,3 @@ def get_free_ports(selected_range, exclude=None, nr=1, client=None):
...
                 return free_ports
         raise ValueError('Unable to find requested nr of free ports')
-
-    @staticmethod
-    def read_config(filename, client=None):
-        """
-        Read a configuration file
-        :param filename: File to read
-        :type filename: str
-
```
JeffreyDevloo, line 183: Removed because deprecated?

ovs/extensions/generic/system.py
```diff
@@ -173,54 +173,3 @@ def get_free_ports(selected_range, exclude=None, nr=1, client=None):
...
-            cp.readfp(StringIO(contents))
-            return cp
-
-    @staticmethod
-    def write_config(config, filename, client=None):
-        """
-        Write a configuration file
-        :param config: Contents to write
-        :type config: RawConfigParser
-
```
JeffreyDevloo, line 208: Removed because deprecated?

JeffreyDevloo (changes requested):

ovs/extensions/generic/system.py
khenderick: Yes

ovs/extensions/generic/system.py
khenderick: No, just didn't like them, they are used like _everywhere_

JeffreyDevloo (approved):

PR #1172, opened 2016-11-17
Make sure the owner and rights of a file-to-overwrite are kept

khenderick:


JeffreyDevloo (approved):
Copying should resolve the permission problems we have faced.

PR #1181, opened 2016-11-21
Make oauth2 implementation more compliant

grimpy:
Pass state variable
Pass arguments as post data (form)
Pass grant_type empty
Do not pass client_secret in headers
Make it posible to configure redirect_uri

### DIFF
--- a/webapps/api/oauth2/redirectview.py
+++ b/webapps/api/oauth2/redirectview.py
@@ -67,14 +67,13 @@ class OAuth2RedirectView(View):
         base_url = Configuration.get('/ovs/framework/webapps|oauth2.token_uri')
         client_id = Configuration.get('/ovs/framework/webapps|oauth2.client_id')
         client_secret = Configuration.get('/ovs/framework/webapps|oauth2.client_secret')
+        redirect_uri = 'https://{0}/api/oauth2/redirect/'.format(request.META['HTTP_HOST'])
         parameters = {'grant_type': 'authorization_code',
-                      'redirect_url': 'https://{0}/api/oauth2/redirect/'.format(System.get_my_storagerouter().ip),
-                      'client_id': client_id,
+                      'redirect_uri': redirect_uri,
+                      'state': state,
                       'code': code}
-        url = '{0}?{1}'.format(base_url, urllib.urlencode(parameters))
-        headers = {'Accept': 'application/json',
-                   'Authorization': 'Basic {0}'.format(base64.b64encode('{0}:{1}'.format(client_id, client_secret)).strip())}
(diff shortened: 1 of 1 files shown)


khenderick:
I'm not sure how this change is in any way compliant with RFC 6749. The changed code is (supposed to be) part of https://tools.ietf.org/html/rfc6749#section-4.1 step D and E but would after this change be invalid.

grimpy:
Fair remarks (altough some of those things nobody does), Ill make some adjustments in the course of this week.

khenderick:
Where exactly do you see non-compliant code, I've double checked the code and I can't see anything that might be missing.

grimpy:
params should (atleast in most common implementations) be passed via post data
request_uri instead of request_url (i instead of l)
the request_uri will moslty likely be some public domain not the ip of the storage node

khenderick:
Makes sense indeed. But the custom configurable ```redirect_uri``` should also be specified in the root of the API since that's used by the first three steps of the flow.

grimpy:
I've modified the changes. Now the only real change is that data is in postdata instead of query params.
And redirect_uri is written correct and passed a sane default.
redirect_uri is actually optional when requesting access_token (in our oauth2 server ignored anyway)

webapps/api/oauth2/redirectview.py
```diff
@@ -67,14 +67,12 @@ def get(self, request, *args, **kwargs):
         base_url = Configuration.get('/ovs/framework/webapps|oauth2.token_uri')
         client_id = Configuration.get('/ovs/framework/webapps|oauth2.client_id')
         client_secret = Configuration.get('/ovs/framework/webapps|oauth2.client_secret')
+        redirect_uri = 'https://{0}/api/oauth2/redirect/'.format(request.META['HTTP_HOST'])
         parameters = {'grant_type': 'authorization_code',
-                      'redirect_url': 'https://{0}/api/oauth2/redirect/'.format(System.get_my_storagerouter().ip),
-                      'client_id': client_id,
```
khenderick, line 72: Shouldn't the ```client_id``` be part of the call cfr https://tools.ietf.org/html/rfc6749#section-4.1.3?

grimpy: REQUIRED, if the client is not authenticating with the
         authorization server as described in Section 3.2.1.

If you jump to seciont 3.2.1 and from there to 2.3 you will see its not required if you are passing the client_secret via authorization header.

khenderick: Aha, ok, thanks

khenderick (approved):

khenderick:
Packaged in openvstorage-2.7.6-rev.4325.b08f4a0

PR #1182, opened 2016-11-21
Handle \u2019 correct and replace any other non-translated unicode

khenderick:
Damn, I ? unicode...

### DIFF
--- a/ovs/extensions/generic/sshclient.py
+++ b/ovs/extensions/generic/sshclient.py
@@ -213,11 +213,14 @@ class SSHClient(object):
             else:
                 cleaned = text.strip().decode('utf-8', 'replace')
             for old, new in {u'\u2018': "'",
+                             u'\u2019': "'",
                              u'\u201a': "'",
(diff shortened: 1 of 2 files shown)


JeffreyDevloo (dismissed):

JeffreyDevloo:
The current implementation is fine as it as however it would be more scalable to rely on python to decode our unicode strings.
For example:
```
In [1]: a=u"aaaàçççñññ"

In [2]: type(a)
Out[2]: unicode

In [3]: a.encode('ascii','ignore')
Out[3]: 'aaa'

In [4]: a.encode('ascii','replace')
Out[4]: 'aaa???????'

In [5]: import unicodedata

In [6]: unicodedata.normalize('NFKD', a).encode('ascii','ignore')
Out[6]: 'aaaacccnnn'

```
The last option with the normalize seems to be ideal as certain characters will get swapped for their 'normal' counterpart.

JeffreyDevloo (approved):
Should allow parsing of all unicode strings.

Issue #1257, opened 2016-12-09
Update of refresh_package_information is not threadsafe


pploegaert:
As observed on OVH environment during update and via code snippet:

```
152 - WARNING - 172.20.20.231: Failed to find a version file in /opt/OpenvStorage/run for service arakoon-voldrv
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "ovs/lib/update.py", line 643, in execute_update
    services_to_start_again=services_stop_start)
  File "ovs/lib/update.py", line 703, in _recover_from_failed_update
    UpdateController._refresh_package_information(local_ip=local_client_ip)
  File "ovs/lib/update.py", line 688, in _refresh_package_information
    GenericController.refresh_package_information()
  File "/usr/lib/python2.7/dist-packages/celery/local.py", line 188, in call
    return self._get_current_object()(*a, **kw)
  File "/usr/lib/python2.7/dist-packages/celery/app/task.py", line 420, in call
    return self.run(*args, **kwargs)
  File "ovs/lib/helpers/decorators.py", line 239, in new_function
    output = function(*args, **kwargs)
  File "ovs/lib/helpers/..//generic.py", line 534, in refresh_package_information
    raise Exception(' - {0}'.format('\n - '.join(errors)))
Exception:  - 172.20.20.103: Authentication failed.

```

Code snippet:

```
from ovs.lib.update import UpdateController
for x in xrange(10):
    UpdateController._refresh_package_information()
```

khenderick:
The code snippet won't work, but I think the issue is that the SSHClient isn't thread-safe, which does make sense. So we should reduce the number of threads used.

kvanhijf:
Code snippet i used in order to reproduce:
```
def test():
    from ovs.lib.generic import GenericController
    import time
    attempts = 50
    for x in range(attempts):
        print x
        start = time.time()
        GenericController.refresh_package_information()
        print time.time() - start
        print ''
```

kvanhijf:
```
In [2]: test()
0
8.5633430481

1
6.89185500145

2
6.80949091911

3
6.84956097603

4
6.79776716232

5
6.94766998291

6
6.7781188488

7
6.74318599701

8
6.72256708145

9
6.78440499306

10
6.78925609589

11
6.76898193359

12
6.7696928978

13
6.82977795601

14
7.02029395103

15
6.86195397377

16
6.91676092148

17
6.83101701736

18
6.99339199066

19
7.13648891449

20
[...]

PR #1261, opened 2016-12-12
#1257 - Made refresh_package_information thread safe

kvanhijf:


ovs/extensions/generic/sshclient.py
```diff
@@ -143,7 +145,11 @@ def __init__(self, endpoint, username='ovs', password=None):
 
         if not self.is_local:
             logging.getLogger('paramiko').setLevel(logging.WARNING)
-            key = '{0}@{1}'.format(self.ip, self.username)
+            if cached is True:
+                key = '{0}@{1}'.format(self.ip, self.username)
+            else:
+                random_name = ''.join(random.choice(string.ascii_letters + string.digits) for _ in range(10))
```
khenderick: I'd rather not cache it at all when no cache is required, to preserve memory.

khenderick (changes requested):

khenderick (approved):

kvanhijf:
openvstorage-2.7.8-rev.4429.a90f470

PR #1265, opened 2016-12-13
Fix race condition during disk creation

khenderick:
Removes possiblity of ending up with two disks in the model

### DIFF
--- a/ovs/lib/vdisk.py
+++ b/ovs/lib/vdisk.py
@@ -296,7 +296,6 @@ class VDiskController(object):
                 new_vdisk = VDiskList.get_by_devicename_and_vpool(devicename, vdisk.vpool)
                 if new_vdisk is None:
                     new_vdisk = VDisk()
-                    new_vdisk.name = VDiskController.extract_volumename(devicename)
                     new_vdisk.volume_id = volume_id
                     new_vdisk.size = vdisk.size
                     new_vdisk.name = name
@@ -562,7 +561,6 @@ class VDiskController(object):
                 new_vdisk = VDiskList.get_by_devicename_and_vpool(devicename, vdisk.vpool)
                 if new_vdisk is None:
                     new_vdisk = VDisk()
-                    new_vdisk.name = VDiskController.extract_volumename(devicename)
                     new_vdisk.volume_id = volume_id
                     new_vdisk.size = vdisk.size
                     new_vdisk.name = name
--- a/ovs/lib/vdisk.py
+++ b/ovs/lib/vdisk.py
@@ -298,10 +298,10 @@ class VDiskController(object):
                     new_vdisk = VDisk()
                     new_vdisk.volume_id = volume_id
                     new_vdisk.size = vdisk.size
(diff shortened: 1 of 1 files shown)


ovs/lib/vdisk.py
```diff
@@ -291,15 +291,20 @@ def clone(vdisk_guid, name, snapshot_id=None, storagerouter_guid=None):
...
-            new_vdisk.size = vdisk.size
+            new_vdisk = VDiskList.get_vdisk_by_volume_id(volume_id)
+            if new_vdisk is None:
+                new_vdisk = VDiskList.get_by_devicename_and_vpool(devicename, vdisk.vpool)
+                if new_vdisk is None:
+                    new_vdisk = VDisk()
+                    new_vdisk.name = VDiskController.extract_volumename(devicename)
+                    new_vdisk.volume_id = volume_id
+                    new_vdisk.size = vdisk.size
+                    new_vdisk.name = name
```
kvanhijf: Saving the name twice might give unexpected results

ovs/lib/vdisk.py
```diff
@@ -552,14 +557,19 @@ def create_from_template(vdisk_guid, name, storagerouter_guid=None):
...
-            new_vdisk.devicename = devicename
+            new_vdisk = VDiskList.get_vdisk_by_volume_id(volume_id)
+            if new_vdisk is None:
+                new_vdisk = VDiskList.get_by_devicename_and_vpool(devicename, vdisk.vpool)
+                if new_vdisk is None:
+                    new_vdisk = VDisk()
+                    new_vdisk.name = VDiskController.extract_volumename(devicename)
+                    new_vdisk.volume_id = volume_id
+                    new_vdisk.size = vdisk.size
+                    new_vdisk.name = name
```
kvanhijf: Same here

ovs/lib/vdisk.py
```diff
@@ -617,14 +627,18 @@ def create_new(volume_name, volume_size, storagedriver_guid):
...
-            new_vdisk.save()
+            new_vdisk = VDiskList.get_vdisk_by_volume_id(volume_id)
+            if new_vdisk is None:
+                new_vdisk = VDiskList.get_by_devicename_and_vpool(devicename, vpool)
+                if new_vdisk is None:
+                    new_vdisk = VDisk()
+                    new_vdisk.size = volume_size
+                    new_vdisk.vpool = vpool
+                    new_vdisk.devicename = devicename
+                    new_vdisk.name = volume_name
```
kvanhijf: No ```extract_volumename``` required here?

kvanhijf (changes requested):

ovs/lib/vdisk.py
khenderick: No, these ```extract_volumename``` code was actually a copy-paste leftover from the ```resize_from_voldrv``` call where - if the disk does not exist yet - it sets a best-guess name.

kvanhijf (approved):

PR #1302, opened 2016-12-23
Multiple storagedrivers

khenderick:


ovs/lib/mdsservice.py
```diff
@@ -94,27 +114,27 @@ def prepare_mds_service(storagerouter, vpool, fresh_only):
...
+        min_load = min(sr_sd_map[storagerouter.name].keys())
+        least_loaded_storagedriver = random.choice(sr_sd_map[storagerouter.name][min_load])
+        if fresh_only is True and min_load > 0:
+            raise RuntimeError('Incorrect MDS service deployment found')
 
         # MODEL UPDATES
-        # 1. Service
-        service_number += 1
+        # Service and MDS service
+        service_number = len(mdsservice_type.services)
```
khenderick: If you would remove an MDS in the middle, this will yield an already existing number

khenderick:
Closed this PR as the branch is too outdated. Will need major rework if to be implemented once more.

Issue #1068, opened 2016-10-20
Avahi reports: Failed to add service, but setup is completed with success.


kinvaris:
When trying to setup my second node, I got the problem that the cluster was not visible. When looking deeper into avahi. It seemed like the character limit was reached for the avahi service.

`Oct 20 13:21:14 ovs-node01-1604 avahi-daemon[26250]: Failed to add service 'ovs_cluster_jonas-test-20okt-1604_ovs-node01-1604`

The character limit for avahi is 65 characters.

This should be checked when trying to enter a new cluster name!

wimpers:
Let's limit the names of the cluster and hostname.

wimpers:
Was earlier reported also http://jira.cloudfounders.com/browse/OVS-4099

kinvaris:
@wimpers, we should also need to cut down on `ovs_cluster` part because it takes up 20% of the string

kvanhijf:
https://github.com/openvstorage/framework/pull/1303 --> openvstorage-2.7.9-rev.4476.91633b0

@QA:
Avahi clusternames are limited to 63 characters
Therefore the implementation is now as following:
```
'ovs_cl_<cluster_name>_<node_name>_<ip>'
```
For \<node_name\> we take the last 24 characters of the name in case its longer than 24 chars
Eg:
node_name = 'my_ovs_cluster_my_long_node_name' --> would become 'luster_my_long_node_name'

For \<ip\> a maximum of 15 chars is preserved (eg: 255.255.255.255)

The start part 'ovs_cl' is 6 chars, so substracting all this (along with the underscores) from the maximum of 63, we end with this:
```
63 - 6 - 15 - 24 - 3 (underscores) = 15
```
So the <cluster_name> can always have a minimum of 15 characters (and potentially longer if the node name is shorter)

Tested scenarios:
* 1st node: New code - 2nd node: New code (SUCCESS)
* 1st node: Old code - 2nd node: New code (SUCCESS)
* 1st node: New code - 2nd node: Old code (FAILURE) --> Can't let nodes with old code join a cluster set up using new code (joining manually instead does work for this)

PR #1303, opened 2016-12-26
#1068 - Validation of Avahi cluster name

kvanhijf:


JeffreyDevloo (approved):

ovs/lib/nodeinstallation.py
```diff
@@ -216,14 +215,19 @@ def setup_node(node_type=None, execute_rollback=False):
                     if cluster_name == new_cluster:  # Create a new OVS cluster
                         first_node = True
                         while True:
+                            master_ip = Interactive.ask_choice(NodeInstallationController.host_ips, 'Select the public IP address of {0}'.format(node_name))
```
pploegaert, line 218: This changes the order of the interactive setup ... pls move the question for the public IP after the cluster_name, as we now can ask to create a new cluster, get a question for the ip and then the question for the cluster_name

This breaks the interactive setup and requires an update of the documentation too ...

wimpers:
@pploegaert  @kvanhijf  

what exactly needs to be changed in the documentation?

pploegaert:
preferably nothing, just reorder the questions in the interactive setup as it was before
otherwise we need to update the nightly build setup too ...

kvanhijf:
@wimpers : We now ask
"Select the public IP address" before "Please enter the cluster name"
This used to be the other way around

I'd rather not revert this, because of my previous explanation.
Either the framework changes the order in which the questions are asked or QA does it.
On the other hand i wonder why QA doesn't use the preconfig.json file for setting up environments like on 16.04, then we wouldn't have these issues. The manual installations are still being verified by the framework team anyway

pploegaert:
@kvanhijf we test both automatic and manual,
when the interactive part fails we can at least check if this is valid or we at least need to update the tests and the documentation
this is more work so what does this change in order bring us?
this is not related to an avahi change anyway, right?

kvanhijf:
It is related to the Avahi cluster name validation, because the IP is part of the internal Avahi cluster name, so if you change IP, the cluster name can become shorter/longer
Also imo, QA should reflect changes made by the framework in their code, the framework should not write code in order to let the tests written by QA succeed

wimpers:
@kvanhijf has 2 points

PR #1346, opened 2017-01-17
Adds the ability to regenerate service files.

JeffreyDevloo:


ovs/extensions/services/systemd.py
```diff
@@ -105,6 +105,41 @@ def add_service(name, client, params=None, target_name=None, startup_dependency=
...
+                    startup_dependency.split('.')[:-1])  # Remove .service from startup dependency
+            output = Systemd.add_service(name=name,
+                                         client=client,
+                                         params=service_params,
+                                         target_name=target_name,
+                                         startup_dependency=startup_dependency,
+                                         delay_registration=True)
+            if output is None:
+                raise RuntimeError('Regenerating files for service {0} has failed'.format(target_name))
+            else:
```
khenderick: Just `return True` here instead of the `else`-branch

khenderick (changes requested):

ovs/extensions/services/systemd.py
```diff
@@ -105,6 +105,41 @@ def add_service(name, client, params=None, target_name=None, startup_dependency=
         return params
 
     @staticmethod
+    def regenerate_service(name, client, target_name):
+        """
```
kvanhijf, line 110: What exactly does this method?

ovs/extensions/services/systemd.py
```diff
@@ -105,6 +105,41 @@ def add_service(name, client, params=None, target_name=None, startup_dependency=
...
+        :param name: Template name of the service to add
```
kvanhijf: You don't add a service

ovs/extensions/services/systemd.py
```diff
@@ -105,6 +105,41 @@ def add_service(name, client, params=None, target_name=None, startup_dependency=
...
+        :type name: str
+        :param client: Client on which to add the service
+        :type client: ovs.extensions.generic.sshclient.SSHClient
+        :param target_name: The current service name eg ovs-volumedriver_flash01.service
+        :type target_name: str
+        :return: True if successful
```
kvanhijf: Shouldn't return anything imo. Raise if anything goes wrong, None in successful scenario

ovs/extensions/services/systemd.py
```diff
@@ -105,6 +105,41 @@ def add_service(name, client, params=None, target_name=None, startup_dependency=
...
+        :rtype bool
+        """
+        configuration_key = '/ovs/framework/hosts/{0}/services/{1}'.format(System.get_my_machine_id(client),
```
kvanhijf: The key `/ovs/framework/hosts...` should be a class variable

ovs/extensions/services/systemd.py
```diff
@@ -105,6 +105,41 @@ def add_service(name, client, params=None, target_name=None, startup_dependency=
...
+                                                                           target_name.replace('ovs-', ''))
+        # If the entry is stored in arakoon, it means the service file was previously made
+        if Configuration.exists(configuration_key):
```
kvanhijf: If key does not exist, an error should be thrown mentioning `add_service` should be used

ovs/extensions/services/upstart.py
```diff
@@ -98,6 +98,41 @@ def add_service(name, client, params=None, target_name=None, startup_dependency=
         return params
 
     @staticmethod
+    def regenerate_service(name, client, target_name):
+        """
+        :param name: Template name of the service to add
```
kvanhijf: Identical remarks as for systemd

kvanhijf (changes requested):

PR #1347, opened 2017-01-17
Remove Etcd from docs

wimpers:


kvanhijf:
These occurrences have been removed by the framework team since its in the framework repository
Closing PR
However these changes are still on the development branch, since the removal of Etcd itself is there too

PR #1356, opened 2017-01-18
Prevent duplicate Domains to be created using the API

khenderick:


webapps/api/backend/views/domains.py
```diff
@@ -72,6 +72,10 @@ def create(self, request, contents=None):
         contents = None if contents is None else contents.split(',')
         serializer = FullSerializer(Domain, contents=contents, instance=Domain(), data=request.DATA)
         domain = serializer.deserialize()
+        current_domains = DomainList.get_by_name(domain.name)
+        if len(current_domains) > 0:
+            raise HttpNotAcceptableException(error_description='A Domain with the given name already exist',
```
kvanhijf: `exists`

ovs/dal/migration/ovsmigrator.py
```diff
@@ -30,7 +30,7 @@ class OVSMigrator(object):
     """
 
     identifier = 'ovs'
-    THIS_VERSION = 14
+    THIS_VERSION = 15
```
kvanhijf: Why bump this when no migration code is present?

kvanhijf (changes requested):

ovs/dal/migration/ovsmigrator.py
khenderick: Added an index, so the migration code for indexes has to run again.

kvanhijf: Yes, but since we're still in development, i guess this isn't required?

khenderick: Hmm, valid point. I'll keep it on 14

kvanhijf (approved):

PR #1396, opened 2017-02-02
Correct event processing

khenderick:


ovs/lib/tests/vdisk_tests/test_events.py
```diff
@@ -0,0 +1,153 @@
...
+        self.persistent.clean()
+        StorageRouterClient.clean()
+
+    def tearDown(self):
+        """
+        Clean up the unittest
+        """
+        # Cleaning storage
+        self.volatile.clean()
+        self.persistent.clean()
```
kvanhijf, line 65: Teardown should also call `StorageRouterClient.clean()`
or a TearDownClass should do so

ovs/lib/tests/vdisk_tests/test_events.py
```diff
@@ -0,0 +1,153 @@
...
+        # * delete volume, delete event,
+        # * create volume, resize event,
+        # * delete volume, delete event
+
+        # Let's test the normal flow
+        first_volume_id = srclient.create_volume(devicename, backend_config, size, storagedriver.storagedriver_id)
+        VDiskController.resize_from_voldrv(first_volume_id, size, devicename, storagedriver.storagedriver_id)
+        self.assertEqual(len(srclient.list_volumes()), 1)
+        self.assertEqual(len(vpool.vdisks), 1)
+        srclient.unlink(devicename)
```
kvanhijf, line 99: why not call `VDiskController.delete()`, this does the unlinking and delete from VolumeDriver for you?

ovs/lib/vdisk.py
```diff
@@ -94,6 +94,24 @@ def clean_vdisk_from_model(vdisk):
         vdisk.delete()
 
     @staticmethod
+    def vdisk_checkup(vdisk):
+        """
+        Triggers a few (async) tasks to make sure the vDisk is in a healthy state.
+        :param vdisk: The vDisk to check
+        :type vdisk: ovs.dal.hybrid.vdisk.VDisk
```
kvanhijf: If you do it, do it good ;)

ovs/lib/vdisk.py
```diff
@@ -104,12 +122,12 @@ def delete_from_voldrv(volume_id):
         :type volume_id: str
         :return: None
         """
-        vdisk = VDiskList.get_vdisk_by_volume_id(volume_id)
-        if vdisk is not None:
-            with volatile_mutex('voldrv_event_disk_{0}'.format(vdisk.volume_id), wait=20):
+        with volatile_mutex('voldrv_event_disk_{0}'.format(volume_id), wait=20):
```
kvanhijf, line 125: 20 seconds to clean the vDisk from the model?  Isn't that a bit much?

kvanhijf (changes requested):

ovs/lib/tests/vdisk_tests/test_events.py
khenderick: Because in this test, I want to have full control about when exactly the "events" are processed. For this specific part, I could indeed do that, but for the below scenarios I want to execute some things in between.

ovs/lib/vdisk.py
khenderick: It is. But this is the time it will wait for the lock. So if e.g. 5 other threads/processes are waiting for the same lock, they add up.

kvanhijf (approved):

PR #1425, opened 2017-02-08
Add API to sync vDisks with reality

khenderick:


ovs/lib/vdisk.py
```diff
@@ -1287,6 +1287,52 @@ def restart(vdisk_guid, force):
...
+                            new_vdisk.name = name
+                            new_vdisk.description = name
+                            new_vdisk.devicename = devicename
+                            new_vdisk.size = new_vdisk.info['volume_size']
+                            new_vdisk.metadata = {'lba_size': new_vdisk.info['lba_size'],
+                                                  'cluster_multiplier': new_vdisk.info['cluster_multiplier']}
+                            new_vdisk.pagecache_ratio = 1.0
+                            new_vdisk.save()
+                            VDiskController.vdisk_checkup(new_vdisk)
+            for volume_id in [volume_id for volume_id in vdisks.keys() if volume_id not in volume_ids]:
```
kvanhijf: Wouldn't it be better to have an `else` stack here to pop the `volume_id` out the `vdisks` dict?
Then this current line is much shorter

kvanhijf (changes requested):

kvanhijf (approved):

PR #1476, opened 2017-02-23
Add timeout support to SSHClient

khenderick:


ovs/extensions/generic/sshclient.py
```diff
@@ -311,14 +334,17 @@ def run(self, command, debug=False, suppress_logging=False, allow_nonzero=False,
...
-            output = self._clean_text(stdout.readlines())
-            error = self._clean_text(stderr.readlines())
-            exit_code = stdout.channel.recv_exit_status()
+            _, stdout, stderr = self._client.exec_command(command, timeout=timeout)  # stdin, stdout, stderr
+            try:
+                output = self._clean_text(stdout.readlines())
+                error = self._clean_text(stderr.readlines())
+                exit_code = stdout.channel.recv_exit_status()
+            except socket.timeout:
+                raise CalledProcessTimeout(124, original_command, 'Timeout during command')
```
kvanhijf, line 347: In the `LocalClient` you use exit_code io `124`

kvanhijf: We also have a `df` command in `storagerouter.py`

kvanhijf (changes requested):

ovs/extensions/generic/sshclient.py
khenderick: The `SSHClient` also uses `exit_code` when it checks it is `124`.

kvanhijf (approved):

PR #1479, opened 2017-02-24
Fixed an issue in the ensure single decorator

kvanhijf:
Eg: 3 tasks shortly after eachother launched (inline), so executed by 3 workers
Task 1 starts processing, task 2 is put on the queue, task 3 was discarded
However task 3 relies on the changes Task 1 or task 2 should do, but since its discarded, it returned immediately
So the fix includes waiting for task 1 & task 2 in case its an inline task

Also made some optimizations in GUI


ovs/lib/helpers/decorators.py
```diff
@@ -114,9 +111,10 @@ def wrap(function):
         Wrapper function
         :param function: Function to check
         """
-        def new_function(*args, **kwargs):
+        def new_function(self=None, *args, **kwargs):
```
khenderick: Can you check whether this should be optional?

ovs/lib/helpers/decorators.py
```diff
@@ -215,12 +217,33 @@ def update_value(key, append, value_to_update=None):
...
+                            if async_task is True:  # Not waiting for other jobs to finish since asynchronously
+                                log_message('Execution of task {0} {1} discarded because of identical parameters'.format(task_name, params_info))
+                                return None
+
+                            # Let's wait for 2nd job in queue to have finished
+                            if persistent_client.exists(persistent_key):
+                                counter = 0
+                                while counter < timeout:
+                                    log_message('Task {0} {1} is waiting for similar tasks to finish - ({2})'.format(task_name, params_info, counter + 1))
+                                    values = persistent_client.get(persistent_key)['values']
```
khenderick: Maybe use `get` with `must_exist` set to `False`, so you can prevent race conditions and act correctly when it might not exist.

ovs/lib/storagedriver.py
```diff
@@ -160,7 +160,7 @@ def cluster_registry_checkup():
         return changed_vpools
 
     @staticmethod
-    @celery.task(name='ovs.storagedriver.scheduled_voldrv_arakoon_checkup', schedule=Schedule(minute='15', hour='*'))
+    @celery.task(name='ovs.storagedriver.scheduled_voldrv_arakoon_checkup', schedule=Schedule(minute='15', hour='*'), bind=True)
     def scheduled_voldrv_arakoon_checkup():
```
khenderick: Might not work

khenderick (changes requested):

ovs/lib/helpers/decorators.py
```diff
@@ -75,7 +75,33 @@ def new_function(*args, **kwargs):
...
+        """
+        Wrapper function
+        """
+        from ovs.celery_run import celery
+
+        ensure_single_info = kwargs.pop('ensure_single_info', {})
+        if ensure_single_info != {} and os.environ.get('RUNNING_UNITTESTS') != 'True':
+            mode = ensure_single_info.pop('mode', None)
+            if mode is None:
+                raise ValueError('Mode is a mandatory keyword argument for the _ensure_single decorator')
```
khenderick: Why manually raise an error while the python runtime would check this for you and raise the correct expected exception instead?

ovs/lib/helpers/decorators.py
```diff
@@ -75,7 +75,33 @@ def new_function(*args, **kwargs):
...
+            f = _ensure_single(task_name=kwargs['name'], mode=str(mode), **ensure_single_info)(f)
+            kwargs['bind'] = True
+        f = celery.task(**kwargs)(f)
```
khenderick: I'd just return this.

ovs/lib/helpers/decorators.py
```diff
@@ -75,7 +75,33 @@ def new_function(*args, **kwargs):
...
 
-def ensure_single(task_name, extra_task_names=None, mode='DEFAULT', global_timeout=300):
+def ovs_task(**kwargs):
+    """
+    Decorator to execute celery tasks in OVS
+    These tasks can be wrapped additionally in the ensure single decorator
+    """
+    if 'name' not in kwargs:
+        raise ValueError('Name is a mandatory keyword argument for the ovs_task decorator')
+
```
khenderick: Wouldn't it be a good idea to add the `@wraps` decorator here, so we don't lose metadata

ovs/lib/storagedriver.py
```diff
@@ -169,7 +167,7 @@ def scheduled_voldrv_arakoon_checkup():
         StorageDriverController._voldrv_arakoon_checkup(False)
 
     @staticmethod
-    @celery.task(name='ovs.storagedriver.manual_voldrv_arakoon_checkup')
+    @ovs_task(name='ovs.storagedriver.manual_voldrv_arakoon_checkup', ensure_single_info={'mode': 'DEFAULT'})
```
khenderick: Will this prevent both `scheduled_voldrv_arakoon_checkup` and `manual_voldrv_arakoon_checkup` to be executed at the same time? I'd expect that they can now run together as their name is different.

khenderick (changes requested):

kvanhijf:
Linked issue: https://github.com/openvstorage/framework/issues/1487

JeffreyDevloo:
Possibly implement https://github.com/openvstorage/framework/issues/1488 too while you are working on it?

ovs/lib/helpers/decorators.py
kvanhijf: No need for the @wraps decorator since we merely merge 2 separate decorators

ovs/lib/helpers/decorators.py
kvanhijf: Adjusted

ovs/lib/helpers/decorators.py
kvanhijf: Adjusted

ovs/lib/storagedriver.py
kvanhijf: Adjusted

ovs/lib/helpers/decorators.py
```diff
@@ -75,7 +75,27 @@ def new_function(*args, **kwargs):
...
+    def wrapper(f):
+        """
+        Wrapper function
+        """
+        from ovs.celery_run import celery
+
+        ensure_single_info = kwargs.pop('ensure_single_info', {})
+        if ensure_single_info != {} and os.environ.get('RUNNING_UNITTESTS') != 'True':
+            mode = ensure_single_info.pop('mode', None)
+            f = _ensure_single(task_name=kwargs['name'], mode=str(mode), **ensure_single_info)(f)
```
khenderick: That might give the string `None`

ovs/lib/helpers/decorators.py
```diff
@@ -172,8 +197,13 @@ def update_value(key, append, value_to_update=None):
...
                         key_to_check = '{0}_{1}'.format(ENSURE_SINGLE_KEY, task)
                         if persistent_client.exists(key_to_check):
-                            log_message('Execution of task {0} discarded'.format(task_name))
-                            return None
+                            if callback is None:
+                                log_message('Execution of task {0} discarded'.format(task_name))
+                                return None
+                            else:
+                                log_message('Execution of task {0} in progress, executing callback function {1}'.format(task_name, callback.__name__))
+                                return callback(*args, **kwargs)
```
khenderick: Does this actually works if you're running the task async on a different node? I'd think it can't reach the callback function?

ovs/lib/helpers/decorators.py
```diff
@@ -192,8 +222,12 @@ def update_value(key, append, value_to_update=None):
                         for task in extra_task_names:
```
khenderick: I'd opt to remove this, we don't use it anyway I think

ovs/lib/helpers/decorators.py
```diff
@@ -215,19 +249,45 @@ def update_value(key, append, value_to_update=None):
...
+                                if callback is None:
+                                    log_message('Execution of task {0} {1} discarded because of identical parameters'.format(task_name, params_info))
+                                    return None
+                                else:
+                                    log_message('Execution of task {0} {1} in progress, executing callback function {2}'.format(task_name, params_info, callback.__name__))
+                                    return callback(*args, **kwargs)
+
+                            # Let's wait for 2nd job in queue to have finished
+                            counter = 0
+                            while counter < timeout:
```
khenderick: Also do a callback if requested.

ovs/lib/helpers/decorators.py
```diff
@@ -215,19 +249,45 @@ def update_value(key, append, value_to_update=None):
...
+                                              'timestamp': now})
 
                 # Poll the arakoon to see whether this call is the only in list, if so --> execute, else wait
                 counter = 0
                 while counter < timeout:
-                    if persistent_client.exists(persistent_key):
-                        values = persistent_client.get(persistent_key)['values']
-                        queued_jobs = [v for v in values if v['kwargs'] == kwargs_dict]
+                    values = list(persistent_client.get_multi([persistent_key], must_exist=False))
+                    if values[0] is not None:
```
khenderick, line 281: You should create an else-statement to make sure it e.g. raises.

ovs/lib/storagedriver.py
```diff
@@ -178,7 +179,7 @@ def manual_voldrv_arakoon_checkup():
         StorageDriverController._voldrv_arakoon_checkup(True)
```
khenderick, line 180: We execute this `manual_voldrv_arakoon_checkup` in the `add_vpool`, so it might be discarded due to a running task and we don't wait for it. We either handle this manually (return `True` here, and check for that value in `add_vpool`), or adapt `ensure_single` to support waiting in default mode.

khenderick (changes requested):

ovs/lib/helpers/decorators.py
kvanhijf: ACK

ovs/lib/helpers/decorators.py
kvanhijf: ACK

ovs/lib/helpers/decorators.py
kvanhijf: ACK

ovs/lib/helpers/decorators.py
kvanhijf: ACK

ovs/lib/helpers/decorators.py
kvanhijf: ACK

ovs/lib/storagedriver.py
kvanhijf: ACK

khenderick (approved):

kvanhijf:
https://github.com/openvstorage/framework/pull/1479 --> openvstorage-2.7.13-dev.1488464044.88ca9e5
https://github.com/openvstorage/framework-alba-plugin/pull/374 --> openvstorage-backend-1.7.13-dev.1488464387.a84fedc

PR #1515, opened 2017-03-10
Fix for the SCO cache calculation when no caching is used

kvanhijf:


khenderick (approved):

kvanhijf:
openvstorage-2.8.0-dev.1489153416.c3efa9d

PR #1561, opened 2017-03-28
Enterprise edition support

khenderick:


ovs/extensions/packages/debian.py
```diff
@@ -30,6 +30,22 @@ class DebianPackage(object):
     _logger = LogHandler.get('update', name='package-manager-debian')
 
     @staticmethod
+    def get_release_name(client=None):
+        """
+        Retrieve currently installed versions of the packages provided (or all if none provided)
```
kvanhijf: Incorrect docstring

ovs/extensions/packages/rpm.py
```diff
@@ -29,6 +29,22 @@ class RpmPackage(object):
     _logger = LogHandler.get('update', name='package-manager-rpm')
 
     @staticmethod
+    def get_release_name(client=None):
+        """
+        Retrieve currently installed versions of the packages provided (or all if none provided)
```
kvanhijf: Incorrect docstring

ovs/lib/update.py
```diff
@@ -89,7 +91,9 @@ def _get_package_information_core(client, package_info):
             binaries = PackageManager.get_binary_versions(client=client, package_names=UpdateController._packages_core['storagedriver'])
             installed = PackageManager.get_installed_versions(client=client, package_names=UpdateController.packages_core_all)
             candidate = PackageManager.get_candidate_versions(client=client, package_names=UpdateController.packages_core_all)
-            if set(installed.keys()) != set(UpdateController.packages_core_all) or set(candidate.keys()) != set(UpdateController.packages_core_all):
+            installed_difference = set(UpdateController.packages_core_all) - set(installed.keys())
+            candidate_difference = set(UpdateController.packages_core_all) - set(candidate.keys())
+            if len(installed_difference | candidate_difference) > 3 or any(['volumedriver' not in package and 'alba' not in package for package in installed_difference | candidate_difference]):
```
kvanhijf: Shouldn't we check here, that as soon as there's a difference, its either all `non-ee` packages or all `ee` packages
Cause with this check, its possible to have a mixture of packages between `ee` and `non-ee` and that's not allowed right?

Same remark in the ASD-manager code then actually

ovs/lib/update.py
```diff
@@ -124,17 +128,25 @@ def _get_package_information_core(client, package_info):
                 storagedriver_services.append('dtl_{0}'.format(sd.vpool.name))
                 storagedriver_services.append('volumedriver_{0}'.format(sd.vpool.name))
 
+            alba_package = 'alba' if 'alba' in installed.keys() else 'alba-ee'
+            volumedriver_base_package = 'volumedriver-no-dedup-base' if 'volumedriver-no-dedup-base' in installed.keys() else 'volumedriver-ee-base'
+            volumedriver_server_package = 'volumedriver-no-dedup-server' if 'volumedriver-no-dedup-server' in installed.keys() else 'volumedriver-ee-server'
+
+            version_mapping = {'volumedriver-server': ['volumedriver-no-dedup-server', 'volumedriver-ee-server'],
```
kvanhijf, line 158: This is expected right?  Twice volumedriver-server package io volumedriver-base? Because i guess we check only the volumedriver-server package at this point?

kvanhijf (changes requested):

ovs/lib/update.py
khenderick: Correct. I'll make a better check (it will be quite some code I guess)

ovs/lib/update.py
khenderick: Yes, it's expected

kvanhijf (dismissed):

kvanhijf (approved):

Issue #1559, opened 2017-03-27
Scrub enhancements


saelbrec:
- [x] Allow and correctly handle scrubbing multiple vdisks(of the same or different vpool) on the same storagerouter simultaneously. Code inspection revealed that we need to further investigate and fix potential issue in the proxy cleanup section of scrub job where the first one upon completion kills the proxy for all other remaining instances.

Depending on the work involved, we can do following items now or later
- [x] API to take as param a list of vdisks
- [x] Make the maximum number of allowed scrub jobs configurable on a per node base.

khenderick:
At this point, all disks to be scrubbed are put in a queue. A number of "stacks" are created (a "stack" being a combination of vPool/StorageRouter) and they are processing this same queue simultaneously. Every stack sets up its Proxy and removes it once the queue is empty. The issue is that if a given stack is executed multiple times, they will conflict each other.

What look to me like the safe options:
* Add threads within a "stack" so a single "stack" can process multiple disks at the same time (with the same Proxy)
* Block the same "stack" to be executed multiple times
* Add an API that can scrub a list of vDisks of a given vPool.

Implications:
* Calling the single-disk-scrub API multiple times will cause them to be executed one-by-one
* Calling the multiple-disk-scrub API will allow usage of the parallel scrubbing.

wimpers:
@khenderick will this be part of RC4?

khenderick:
No, it will be part of Gilbert 2.8.1.

kvanhijf:
openvstorage-2.8.1-dev.1490966663.561c40e

@openvstorage/qa : 
New API call on vPools: scrub_multiple_vdisks. Takes a list of vdisk_guids as parameter and the vpool they belong to. This allows a user to scrub multiple vdisks simultaneously.

Also the amount of simultaneous scrub jobs per SCRUB partition can now be configured (default 2)
This could proof useful for dedicated SCRUB nodes
To configure it, do the following:
`Configuration.set(key='/ovs/framework/hosts/<machine_id>/config|scrub_stack_threads', value=5)`

Values between 1 and 20 are enforced, so anything falling out of that range will be overruled internally

# How scrubbing distribution works:

- Collect StorageRouters with SCRUB role
- Make buckets with vDisks per vPool (Queues)
- Launch calculated amount of threads to do the scrubbing on the different StorageRouters, each thread will do:
  - Spawn a scrub proxy
  - Deploy several other threads (configurable, but 2 is default), to execute the actual scrubbing
[...]

PR #1564, opened 2017-03-30
#1559 - Scrub enhancements

kvanhijf:
New API to scrub multiple vDisks of single vPool


ovs/lib/generic.py
```diff
@@ -391,6 +433,14 @@ def _verify_mds_config(current_vdisk):
                                 GenericController._logger.warning('Scrubber - vPool {0} - StorageRouter {1} - vDisk {2} - Skipping because master MDS still not local'.format(vpool.name, storagerouter.name, vdisk.name))
                                 continue
 
+                        # Check if vDisk is already being scrubbed
+                        with volatile_mutex('execute_scrub_work', wait=5):
+                            volatile_key = 'scrub_vdisk_{0}'.format(vdisk_guid)
+                            if volatile_client.get(volatile_key) is not None:
+                                GenericController._logger.warning('Scrubber - vPool {0} - StorageRouter {1} - vDisk {2} - Skipping because vDisk is already being scrubbed'.format(vpool.name, storagerouter.name, vdisk.name))
+                                continue
+                            volatile_client.set(key=volatile_key, value=volatile_key, time=24 * 60 * 60)
```
khenderick: You might be able to get rid of the lock when using `volatile_client.add(...)` instead. This call returns a `False`-ish value when the key already exists. If it doesn't, it adds the key. This is atomic.

ovs/lib/generic.py
```diff
@@ -391,6 +433,14 @@ def _verify_mds_config(current_vdisk):
                                 GenericController._logger.warning('Scrubber - vPool {0} - StorageRouter {1} - vDisk {2} - Skipping because master MDS still not local'.format(vpool.name, storagerouter.name, vdisk.name))
                                 continue
 
+                        # Check if vDisk is already being scrubbed
+                        with volatile_mutex('execute_scrub_work', wait=5):
+                            volatile_key = 'scrub_vdisk_{0}'.format(vdisk_guid)
```
khenderick: Let's add `ovs_` to the key, just in case the memcache is used for some other purpose as well.

khenderick (changes requested):

khenderick (approved):

kvanhijf:
openvstorage-2.8.1-dev.1490881837.17934a9

PR #1587, opened 2017-04-12
Add feature detection for voldrv and alba

khenderick:


ovs/dal/hybrids/storagerouter.py
```diff
@@ -145,3 +148,31 @@ def _recovery_domains(self):
         :return: List of domain guids
         """
         return [junction.domain_guid for junction in self.domains if junction.backup is True]
+
+    def _features(self):
+        """
+        Returns information about installed/available features
+        :return: Dictionary containing edition and available features per component
+        """
+        client = SSHClient(self, username='root')
```
kvanhijf: Retrieving offline StorageRouters or StorageRouters on which the heartbeat data is old, will suffer from this line

ovs/dal/hybrids/storagerouter.py
```diff
@@ -145,3 +148,31 @@ def _recovery_domains(self):
...
+        enterprise_regex = re.compile('^(?P<edition>ee-)?(?P<version>.*)$')
+        version = client.run("volumedriver_fs --version | grep version: | awk '{print $2}'", allow_insecure=True, allow_nonzero=True)
+        volumedriver_version = enterprise_regex.match(version).groupdict()
```
kvanhijf: Is the volumedriver version of the Enterprise edition something like 'ee-6.15.0' ??

kvanhijf (changes requested):

ovs/dal/hybrids/storagerouter.py
khenderick: Correct.

kvanhijf (approved):

PR #1595, opened 2017-04-20
Update metadatastore_bits environment variable from 6 to 5

kvanhijf:


config/templates/systemd/ovs-volumedriver.service
```diff
@@ -5,7 +5,7 @@ Wants=ovs-watcher-volumedriver.service
 
 [Service]
 Type=simple
-Environment=METADATASTORE_BITS=6
+Environment=METADATASTORE_BITS=<METADATA_STORE_BITS>
```
khenderick: Can you keep these equal? The extra `_` is fancy, but a bit confusing.

ovs/lib/migration.py
```diff
@@ -218,7 +220,32 @@ def migrate():
...
+                    entries = ServiceManager.extract_from_service_file(name='ovs-volumedriver_{0}'.format(vpool.name),
+                                                                       client=sr_client_map[storagedriver.storagerouter_guid],
+                                                                       entries=['METADATASTORE_BITS='])
+                    if len(entries) == 1:
+                        bits = entries[0].split('=')[-1]
+                        bits = int(bits) if bits.isdigit() else 5
+
+                if bits is not None:
+                    try:
+                        key = '/ovs/framework/hosts/{0}/services/volumedriver_{1}'.format(storagedriver.storagerouter.machine_id, vpool.name)
```
khenderick: I'd use this logic to check whether the migration still has to be execute. And only if this entry is missing from the contents, you can parse the service file and update the contents (and store it to the `vPool`'s property.

khenderick (changes requested):

khenderick (approved):

Issue #1555, opened 2017-03-24
DTL status is not correct


jeroenmaelbrancke:
There is a problem with the dtl_status.
The DTL status on different components are differents.

volumedriver status:
```
In [3]: vd.name
Out[3]: u'nvmecor3-disk6'

In [4]: vd.info
Out[4]: 
{'cluster_cache_handle': {},
 'cluster_cache_limit': None,
 'cluster_multiplier': 8,
 'failover_ip': '',
 'failover_mode': 'OK_STANDALONE',
```

framework status:
```
In [5]: vd.dtl_status
Out[5]: 'disabled'
```
while healthcheck reported checkup_required (same method as above)
VDisk nvmecor3-disk6s DTL has an unknown status: checkup_required

GUI status:
```
<div class="tooltip-inner">DTL should be configured</div>
```

kvanhijf:
@openvstorage/qa : No idea how to reproduce this specific scenario, but we did make changes to DTL allocation
DTL allocation rules:
* First priority to StorageRouters located in the vDisk's StorageRouter's Recovery Domain
* Second priority to StorageRouters located in the vDisk's StorageRouter's Regular Domain
* If Domains configured, but no StorageRouters are found matching any of the Domains on the vDisk's StorageRouter, a random SR in the same vPool is chosen
* If no Domains configured on the vDisk StorageRouter, any other StorageRouter on which the vPool has been extended is chosen


https://github.com/openvstorage/framework/pull/1634 --> openvstorage-2.9.3-dev.1494405832.d3fd1fc
https://github.com/openvstorage/alba-asdmanager/pull/249 --> openvstorage-sdm-1.9.0-dev.1494406013.e355601
https://github.com/openvstorage/framework-alba-plugin/pull/406 --> openvstorage-backend-1.9.0-dev.1494405928.789193e

PR #1634, opened 2017-05-09
#1555 - DTL status incorrect in certain use-cases

kvanhijf:


ovs/dal/hybrids/vdisk.py
```diff
@@ -107,33 +108,67 @@ def _dtl_status(self):
...
             return 'disabled'
 
-        domains = []
-        possible_dtl_targets = set()
-        for sr in StorageRouterList.get_storagerouters():
-            if sr.guid == self.storagerouter_guid:
-                domains = [junction.domain for junction in sr.domains]
-            elif len(sr.storagedrivers) > 0:
-                possible_dtl_targets.add(sr)
+        self.invalidate_dynamics('storagerouter_guid')
```
khenderick: Just use the `_storagerouter_guid()` instead.

ovs/dal/tests/helpers.py
```diff
@@ -425,10 +427,16 @@ def _set_vpool_storage_driver_configuration(vpool, storagedriver):
...
-        Configuration.set(key, json.dumps(default_config), raw=True)
-        LocalStorageRouterClient.configurations[key] = default_config
+        sds = vpool.storagedrivers
+        if storagedriver is not None:
+            sds = [storagedriver]
+        for sd in sds:
+            key = '/ovs/vpools/{0}/hosts/{1}/config'.format(vpool.guid, sd.storagedriver_id)
+            Configuration.set(key, json.dumps(default_config), raw=True)
+            default_config['volume_router']['vrouter_id'] = sd.storagedriver_id
+            LocalStorageRouterClient.configurations[key] = default_config
```
khenderick: Shouldn't this take a `copy.deepcopy(...)` of the dict, as now you'll always update the `default_config` with another storagedriver_id.

ovs/lib/tests/vdisk_tests/test_dtl_checkup.py
```diff
@@ -16,6 +16,12 @@
 
 """
 DTL checkup test module
+DTL allocation rules:
+    - Priority to StorageRouters which have the recovery Domain of the vDisk StorageRouter, as regular Domain
```
khenderick: A `StorageRouter` is in the domain set as it's primary/regular domain. So wouldn't this be easier if written like "Priority to StorageRouters located in the vDisk's StorageRouter's recovery domain"?

khenderick (changes requested):

ovs/dal/hybrids/vdisk.py
kvanhijf: Yes, but isn't that the purpose of the invalidate? Otherwise we can replace all `invalidate_dynamics` logic with `_<property>()` and thats not how its meant to be i guess?

khenderick: If outside the instance one should indeed not call the hidden function, but inside the class itself I don't see why not to do that if you need the latest version anyway.

kvanhijf: ah, its in the DAL itself, didn't pay enough attention. My bad

khenderick (approved):

kvanhijf:
https://github.com/openvstorage/framework/pull/1634 --> openvstorage-2.9.3-dev.1494405832.d3fd1fc
https://github.com/openvstorage/alba-asdmanager/pull/249 --> openvstorage-sdm-1.9.0-dev.1494406013.e355601
https://github.com/openvstorage/framework-alba-plugin/pull/406 --> openvstorage-backend-1.9.0-dev.1494405928.789193e

PR #1637, opened 2017-05-15
Integrated cache quota

khenderick:


ovs/lib/vdisk.py
```diff
@@ -113,7 +115,48 @@ def vdisk_checkup(vdisk):
...
+                raw_result = client.run(command)
+                try:
+                    results = json.loads(raw_result)
+                except ValueError:
+                    VDiskController._logger.debug('Could not parse result: {0}'.format(raw_result))
+                    raise
+                if results['success'] is False:
+                    raise RuntimeError('Could not set Fragment/Block Cache Quota: {0}'.format(results))
+        except Exception:
+            VDiskController._logger.debug('Executed command: {0}'.format(command))
```
JeffreyDevloo: A command is potentially not executed at this point. Might give off a weird impression?
The command is set prior to its execution. Perhaps set it to None initially and check before logging the debug?

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

PR #1652, opened 2017-06-01
Speedup

khenderick:


webapps/api/backend/decorators.py
```diff
@@ -81,7 +82,11 @@ def new_function(*args, **kw):
             if not ApiToolbox.is_token_in_roles(request.token, roles):
                 raise HttpForbiddenException(error_description='This call requires roles: {0}'.format(', '.join(roles)),
                                              error='invalid_roles')
-            return f(*args, **kw)
+            duration = time.time() - start
```
kvanhijf, line 85: Shouldn't the duration be calculated after having executed the function?

webapps/api/backend/decorators.py
```diff
@@ -215,8 +221,12 @@ def new_function(*args, **kwargs):
             # Execute validator
             if validator is not None:
                 validator(args[0], **validation_new_kwargs)
+            duration = time.time() - start
             # Call the function
-            return f(args[0], **new_kwargs)
+            result = f(args[0], **new_kwargs)
```
kvanhijf, line 226: Same remark

webapps/api/helpers.py
```diff
@@ -0,0 +1,40 @@
...
+"""
+Some helpers
+"""
+
+from rest_framework.response import Response
+
+
+class OVSResponse(Response):
+
+    def __init__(self, data=None, status=200,
```
kvanhijf, line 27: Using a generic approach here, seems better to me
`super(OVSResponse, self).__init__(**kwargs)`

kvanhijf (changes requested):

webapps/api/backend/decorators.py
khenderick: No, as the goal is to time how long the security checking takes, not how long security checking *and* executing the function takes.

webapps/api/helpers.py
khenderick: Would be a bit more awkward since the `timings` arg is at the end, so I'd have to find it in `kwargs` or `args`.

kvanhijf (approved):

PR #1672, opened 2017-06-13
Increase timeout

khenderick:


kvanhijf (approved):

webapps/frontend/lib/ovs/api.js
```diff
@@ -34,7 +34,7 @@ define([
 
         callData = {
             type: type,
-            timeout: generic.tryGet(options, 'timeout', 1000 * 30),
+            timeout: generic.tryGet(options, 'timeout', 1000 * 120),
```
JeffreyDevloo, line 37: I'd suggest a little longer than 120s. The arakoon calls can take up to 120s to timeout alone, leave some room for additional computing time.

JeffreyDevloo (changes requested):

webapps/frontend/lib/ovs/api.js
khenderick: I'd rather keep that down and decrease arakoon timeouts.

JeffreyDevloo: Alright. I'm awaiting a merge in the pyrakoon to be able to overrule the timeouts.

JeffreyDevloo (approved):

PR #1676, opened 2017-06-13
Use RMQ iso memcached as result backend

khenderick:


JeffreyDevloo:
Is rabbitmq a valid resultbackend?

khenderick:
Well, turns out it should not be used. One should either use a cache backend (e.g. memcached) or a "real database".

khenderick:
Seems RMQ is a no-go after all. Closing.

PR #1677, opened 2017-06-13
Use the timeout on the SSHClient to avoid getting stuck

JeffreyDevloo:


ovs/lib/noderemoval.py
```diff
@@ -93,8 +93,8 @@ def remove_node(node_ip, silent=None):
             storage_router_to_remove_online = True
             for storage_router in storage_router_all:
                 try:
-                    client = SSHClient(storage_router, username='root')
-                except (UnableToConnectException, NotAuthenticatedException):
+                    client = SSHClient(storage_router, username='root', timeout=10)
+                except (UnableToConnectException, NotAuthenticatedException, TimeOutException):
```
kvanhijf: Are we sure at this point that the node that times out is the node that went down?
If not, we'll introduce completely different behaviour, because the node which isn't really down, but really slow in responding, will not be unconfigured properly because our logic assumes its down

kvanhijf (changes requested):

ovs/lib/noderemoval.py
kvanhijf: Proposal: Shuffle responsibility to customer. Let him verify the `offline` nodes and show the reasons why they are detected as offline

JeffreyDevloo:
Currently looks like:
```
+++++++++++++++++++++
+++  Remove node  +++
+++++++++++++++++++++
WARNING: Some of these steps may take a very long time, please check the logs for more information


Creating SSH connections to remaining master nodes
  Node with IP 10.100.199.191  successfully connected to
  Unable to connect to node with IP 10.100.199.192 
  Node with IP 10.100.199.193  successfully connected to
  Node with IP 10.100.199.194  successfully connected to

+++ Running "noderemoval - validate_removal" hooks +++

Executing alba._validate_removal
Certain nodes appear to be offline. These will not fully removed and will cause issues if they are not really offline.
Offline nodes: 
  * 10.100.199.192 - Unable to connect.
Continue the removal with these being presumably offline? (y/[n]):
```

ovs/lib/noderemoval.py
```diff
@@ -93,9 +94,16 @@ def remove_node(node_ip, silent=None):
             storage_router_to_remove_online = True
             for storage_router in storage_router_all:
                 try:
-                    client = SSHClient(storage_router, username='root')
-                except (UnableToConnectException, NotAuthenticatedException):
-                    Toolbox.log(logger=NodeRemovalController._logger, messages='  Node with IP {0:<15} is unreachable'.format(storage_router.ip))
+                    client = SSHClient(storage_router, username='root', timeout=10)
+                except (UnableToConnectException, NotAuthenticatedException, TimeOutException) as ex:
+                    if isinstance(ex, UnableToConnectException):
+                        msg = 'Unable to connect'.format(storage_router.ip)
```
khenderick: There's no `{0}` or sth in the string.

khenderick (changes requested):

khenderick (approved):

kvanhijf (approved):

PR #1732, opened 2017-08-24
Http2 instead of http1.1

JeffreyDevloo:
(Seems to building on https://github.com/openvstorage/framework/pull/1731)


kvanhijf (approved):

domsj:
so now it server both http1 & http2, or only http2?

JeffreyDevloo:
Only http2
Http1.1 will still be used for the initial connect to the server on port 80 (no encryption) but it will get redirected to port 443 and using http2

PR #1762, opened 2017-09-11
Added model clean up when something would go wrong during MDS creation

kvanhijf:
Added logging
General clean up

### DIFF
--- a/ovs/lib/mdsservice.py
+++ b/ovs/lib/mdsservice.py
@@ -17,6 +17,7 @@
 """
 MDSService module
 """
+
 import math
 import time
 import random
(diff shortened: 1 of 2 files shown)


ovs/lib/mdsservice.py
```diff
@@ -156,92 +160,107 @@ def prepare_mds_service(storagerouter, vpool, fresh_only):
...
         :type mds_service: ovs.dal.hybrids.j_mdsservice.MDSService
-        :param vpool: The vPool for which the MDS service will be removed
-        :type vpool: ovs.dal.hybrids.vpool.VPool
         :param reconfigure: Indicates whether reconfiguration is required
         :type reconfigure: bool
         :param allow_offline: Indicates whether it's OK that the node for which mds services are cleaned is offline
         :type allow_offline: bool
+        :raises: RuntimeError - When vDisks present on the MDSService to be removed
+                              - No StorageDriver is linked to the MDSService to be removed
+                 UnableToConnectException - When StorageRouter on which the MDSService resides is unreachable and allow_offline flag is False
```
JeffreyDevloo, line 190: A new :raises: might be good for every exception

ovs/lib/mdsservice.py
```diff
@@ -156,92 +160,107 @@ def prepare_mds_service(storagerouter, vpool, fresh_only):
...
+        :rtype: NoneType
         """
-        if mds_service.guid not in vpool.mds_services_guids:
-            raise RuntimeError('MDS Service {0} is not related to vPool {1}'.format(mds_service.service.name, vpool.name))
-
         if len(mds_service.vdisks_guids) > 0 and allow_offline is False:
             raise RuntimeError('Cannot remove MDSService that is still serving disks')
 
         if len(mds_service.storagedriver_partitions) == 0 or mds_service.storagedriver_partitions[0].storagedriver is None:
             raise RuntimeError('Failed to retrieve the linked StorageDriver to this MDS Service {0}'.format(mds_service.service.name))
```
JeffreyDevloo, line 198: On Nuvolat we had the case where the Storagedriver was removed and the mds_service remained in the model.
This way, the service could not get cleanup up in conventional ways.
Not sure if we should tackle with removing our check here though.

ovs/lib/mdsservice.py
```diff
@@ -156,92 +160,107 @@ def prepare_mds_service(storagerouter, vpool, fresh_only):
...
+            MDSServiceController._logger.debug('Remove MDS: StorageRouter {0} - vPool {1}: Established SSH connection'.format(storagerouter.name, vpool.name))
+        except UnableToConnectException:
+            if allow_offline is True:
+                MDSServiceController._logger.warning('Remove MDS: StorageRouter {0} - vPool {1}: Allowed offline node during MDS service removal'.format(storagerouter.name, vpool.name))
+            else:
+                MDSServiceController._logger.exception('Remove MDS: StorageRouter {0} - vPool {1}: Failed to connect to StorageRouter'.format(storagerouter.name, vpool.name))
+                raise
+
+        # Reconfigure StorageDriver
+        if reconfigure is True and root_client is not None:
```
JeffreyDevloo, line 216: Raise on reconfigure is True and root_client is None. The request action could not be done  so we should error here

JeffreyDevloo (changes requested):

ovs/lib/mdsservice.py
kvanhijf: root_client will be None only when allow_offline is True. When allow_offline is True and the StorageRouter is indeed offline, we can't reconfigure anyway.
If allow_offline would be False and the StorageRouter cannot be reached, we raise in the else stack

ovs/lib/mdsservice.py
kvanhijf: Removing a StorageDriver should first check whether above conditions are met before actually doing anything. These checks should however remain in remove_mds_service imo

JeffreyDevloo (approved):

kvanhijf:
openvstorage-2.10.0-dev.1505205363.8328138

PR #1763, opened 2017-09-11
MDS safety granularity per vPool

kvanhijf:


ovs/lib/migration.py
```diff
@@ -245,5 +246,17 @@ def migrate():
                 vpool.metadata_store_bits = bits
                 vpool.save()
 
+        #######################################
+        # MDS safety granularity on vPool level
+        mds_safety_key = '/ovs/framework/storagedriver'
```
JeffreyDevloo: Having this part out of band is risky. The mds checkup could be scheduled and expects these keys while the keys could not yet be set.

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

kvanhijf:
openvstorage-2.10.0-dev.1505202385.ca2d1eb

PR #1767, opened 2017-09-12
Sync vdisks before shrinking

JeffreyDevloo:


ovs/lib/storagerouter.py
```diff
@@ -1086,6 +1086,8 @@ def remove_storagedriver(storagedriver_guid, offline_storage_router_guids=None):
         if client is None:
             raise RuntimeError('Could not find any responsive node in the cluster')
 
+        # Sync with reality to have a clear vision of vDisks
+        VDiskController.sync_with_reality(storage_driver.vpool)
```
kvanhijf: sync_with_reality expects a vpool_guid
And you can use `vpool.guid` i/o `storage_driver.vpool_guid`

kvanhijf (changes requested):

kvanhijf (approved):

PR #1781, opened 2017-09-21
vPool shrink failed due to "Operation canceled" on MDS client(s)

kvanhijf:


ovs/lib/mdsservice.py
```diff
@@ -672,8 +677,10 @@ def _add_suitable_nodes(local_importance, local_safety, services_to_recycle=list
         # Master must be local
         # Master cannot be overloaded
         # Master must be in primary domain (if no domains available, this check is irrelevant because all StorageRouters will match)
+        timeout = 300
```
JeffreyDevloo: Let's introduce an option to overrule this? Timeout section in config management? Would be cool for documentation!

ovs/lib/mdsservice.py
```diff
@@ -890,13 +913,10 @@ def _add_suitable_nodes(local_importance, local_safety, services_to_recycle=list
 
             for service in new_services[1:]:
                 client = mds_client_cache[service]
-                try:
```
JeffreyDevloo: Keep the try, log and raise
This way the ops guys could parse the critical log

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

PR #1785, opened 2017-09-22
Implementation of max tlogs behind functionality

kvanhijf:
WARNING:  Cannot be merged in as long as volumedriver has not ported the functionality to the EE repos too

### DIFF
--- a/ovs/extensions/migration/migration/ovsmigrator.py
+++ b/ovs/extensions/migration/migration/ovsmigrator.py
@@ -66,6 +66,8 @@ class OVSMigrator(object):
             mds_safety_key = '/ovs/framework/storagedriver'
             if Configuration.exists(key=mds_safety_key):
                 current_mds_settings = Configuration.get(key=mds_safety_key)
+                if 'mds_tlogs' in current_mds_settings:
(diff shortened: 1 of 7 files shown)


ovs/extensions/migration/migration/ovsmigrator.py
```diff
@@ -66,6 +66,8 @@ def migrate(previous_version, master_ips=None, extra_ips=None):
             mds_safety_key = '/ovs/framework/storagedriver'
             if Configuration.exists(key=mds_safety_key):
                 current_mds_settings = Configuration.get(key=mds_safety_key)
+                if 'mds_tlogs' in current_mds_settings:
```
JeffreyDevloo, line 69: Reason to popping here?

JeffreyDevloo (changes requested):

ovs/extensions/migration/migration/ovsmigrator.py
kvanhijf: mds_tlogs is configurable via the GUI on the StorageDriver, so no need for this value in our config mgmt, since its available on SD level

JeffreyDevloo (approved):

PR #1790, opened 2017-09-25
Delete snapshot(s) should wait longer when retrieving the snapshots from volumedriver

kvanhijf:


JeffreyDevloo (changes requested):
Let's keep this tidy?
Create a method to call the logic to retrieve the snapshots. Capture the exception in the hybrids as they may not raise exceptions

ovs/lib/vdisk.py
```diff
@@ -459,6 +459,22 @@ def clone(vdisk_guid, name, snapshot_id=None, storagerouter_guid=None, pagecache
...
+        :param vdisk: vDisk to retrieve the snapshot IDs for
+        :type vdisk: ovs.dal.hybrids.vdisk.VDisk
+        :return: The snapshot IDs for the given vDisk
+        :rtype: list
+        """
+        volume_id = str(vdisk.volume_id)
+        try:
+            return vdisk.storagedriver_client.list_snapshots(volume_id, req_timeout_secs=10)
+        except VolumeRestartInProgressException:
+            time.sleep(0.5)
```
JeffreyDevloo, line 474: A while might be better here if we really want to return the snapshot ids when a volume is done restarting (easily done by pointing it to itself)

JeffreyDevloo (approved):

PR #1796, opened 2017-09-26
ignore disks when /dev/<diskname> does not exist

dejonghb:
In lxd containers lsblk leaks info about disks that are not reachable from inside the container (/dev/<disk_name> does not exist). This extra test skips those disks so that the GUI won't show them any more.

### DIFF
--- a/ovs/lib/disk.py
+++ b/ovs/lib/disk.py
@@ -17,6 +17,7 @@
 """
 DiskController module
 """
+import os
 import re
 import time
 from ovs.dal.hybrids.disk import Disk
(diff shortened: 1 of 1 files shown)


ovs/lib/disk.py
```diff
@@ -108,6 +109,9 @@ def sync_with_reality(storagerouter_guid):
             link = client.file_read_link('/sys/block/{0}'.format(name))
             device_state = None
             friendly_path = '/dev/{0}'.format(name)
+            if not os.path.exists(friendly_path):
```
kvanhijf: Should use `client.path_exists(friendly_path)` instead

kvanhijf (changes requested):

kvanhijf (dismissed):

kvanhijf (approved):

kvanhijf:
Package for this PR: openvstorage-2.11.0-dev.1506427655.6219746
Related change: https://github.com/openvstorage/framework-extensions/pull/30 packaged in: openvstorage-extensions-0.3.0-dev.1506427714.a63324d

PR #1812, opened 2017-10-03
no model changes for LXD_FAKE disks

dejonghb:
hack for skipping disks created with model == 'LXD_FAKE' for use in lxd containers

### DIFF
--- a/ovs/lib/disk.py
+++ b/ovs/lib/disk.py
@@ -164,6 +164,8 @@ class DiskController(object):
 
         # Sync the model
         for disk in storagerouter.disks:
+            if disk.model == 'LXD_FAKE':
+                continue
             disk_info = None
             for alias in disk.aliases:
                 if alias in alias_name_mapping:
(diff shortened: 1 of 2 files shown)


JeffreyDevloo (changes requested):
This is a good temporary solution but something better could be thought out

dejonghb:
Maybe these disks could get some extra hint they were "manual" created instead of "autodetected"? Using the model for this is certainly a quick hack that can be improved.
This might also help for  #1814 where the modelled size can/should then overrule what the directory/device underneath reports (in case that would be larger ;-) )?

domsj:
While setting up ovs inside an lxd container I encountered the following error:
https://github.com/openvstorage/framework/blob/develop/ovs/lib/storagerouter.py#L334

The is_mounted check failed, because there is nothing to be mounted on the fakedisk in my test setup.

Perhaps for manually added disks it makes more sense to check whether some marker file is there?

sjovdnbos:
Closed due to deviation being too far from current current branch.

Issue #1687, opened 2017-06-23
Optimize settings for sustainable write throughput


saelbrec:
Tests have shown that the currently configured default settings on a vpool do not allow to guarantee a sustainable write throughput. In certain tests it quickly degrades over a time window of 24hrs.
The following graph shows on the left half test results with the default values, where as the right half shows test results with optimized volumedriver config values for **number_of_scos_in_tlog** and **non_disposable_scos_factor**

![write_test](https://user-images.githubusercontent.com/12733895/27472925-912911fa-57fd-11e7-860f-f3305a72c879.png)

At vpool creation time there are 3 factors to take into account.

1. **SCOsize** can be (4,8,16,32,64,128 MiB)
2. **number_of_scos_in_tlog** has a default of 4 and is recalculated behind the scenes.
3. **non_disposable_scos_factor**, decided upon behind the scenes based on the chosen volume writebuffer size.
Formula: `volume writebuffer = SCOsize * number_of_scos_in_tlog * non_disposable_scos_factor`

Constraints we should implement to make sure we configure better balanced values:

- Enforce non_disposable_scos_factor >= 1,5.
If smaller advise to either decrease SCOsize or increase volume writebuffer as a value to close to 1 would lead to choking.
- If based on `volume writebuffer, SCOsize and number_of_scos_in_tlog(default=4)` the non_disposable_scos_factor > 4.0 , we should increase the number_of_scos_in_tlog to maximum the equivalent of 256MiB of data(256MiB/SCOSizeInMiB=number_of_scos_in_tlog) and then use that value to recalculate the non_disposable_scos_factor.

wimpers:
I'd like to propose a suggestion to the above: use an easy mode and an advanced mode? 

* Easy mode: user defines the write buffer size. We derive/calculate some sane values based upon the write buffer size.
* Advanced mode: User needs to enter write buffer size , SCO size, number_of_scos_in_tlog and non_disposable_scos_factor. We enforce whatever is needed like we do when creating policies.

khenderick:
I kinda like @wimpers his idea. The easy mode asks a few questions and generates a set of works-for-most-setups parameters. The expert mode just exposed (almost) all configuration parameters. This way we can stop changing the default values every month.

sks126:
@wimpers  The following formula in 256 MB is what mean？
                     "256MiB/SCOSizeInMiB=number_of_scos_in_tlog".

ngugc:
[...]

Issue #1678, opened 2017-06-13
Re-use connection for fragment cache when extending vpool


matthiasdeblock:
openvstorage 2.9.5-1

When extending a vpool you do not have the opportunity to re-use a connection you have used before on the fragment cache. 

The block cache does have this option:
![screenshot_2017-06-09_10-21-38](https://user-images.githubusercontent.com/14923932/27086640-0dd558f6-5053-11e7-8e70-3bc9a830caac.png)

It would be very useful to have this for the fragment cache as well.

wimpers:
@matthiasdeblock just to make sure, this is just to make the OPS life easier. You can configure everything correct as needed, right?

matthiasdeblock:
Yes, everything works but it's just a lot of copy/paste work...

JeffreyDevloo:
ლ(ಠ益ಠლ) use the api client to avoid repetitive tasks :P

JeffreyDevloo:
Fixed by 
- #1821 
- https://github.com/openvstorage/framework-iscsi-plugin/pull/34
- https://github.com/openvstorage/framework-extensions/pull/63
- https://github.com/openvstorage/framework-alba-plugin/pull/534

Packaged in:
 - openvstorage_2.12.0-dev.1518011126.4cee8bc-1_amd64.deb
 - openvstorage-backend_1.12.0-dev.1518011188.e274a3d-1_amd64.deb
 - openvstorage-extensions_0.4.0-dev.1518010323.90e5a68-1_amd64.deb
 - openvstorage-iscsi-plugin_0.4.0-dev.1518011220.53302ad-1_amd64.deb

JeffreyDevloo:
Released in https://github.com/openvstorage/framework/releases/tag/2.12.0
Packaged in openvstorage-core_2.12.0-1_amd64.deb

Issue #1669, opened 2017-06-13
TypeError features when creating a vpool


jeroenmaelbrancke:
Changing the storagerouter when creating a vpool throws following error in the devtools.
I wasn't able to select a part of the storagerouters.

```
Uncaught TypeError: Cannot read property 'features' of undefined
    at gather_fragment_cache.js?version=2.9.5:135
    at l (knockout-3.3.0.js?version=2.9.5:44)
    at a.Ub.h [as La] (knockout-3.3.0.js?version=2.9.5:43)
    at Function.notifySubscribers (knockout-3.3.0.js?version=2.9.5:33)
    at Function.d.W (knockout-3.3.0.js?version=2.9.5:36)
    at Object.d [as target] (knockout-3.3.0.js?version=2.9.5:36)
    at Object.self.set (viewmodel.js?version=2.9.5:72)
    at Object.eval (eval at parseBindingsString (knockout-3.3.0.js?version=2.9.5:61), <anonymous>:6:565)
    at HTMLLIElement.<anonymous> (knockout-3.3.0.js?version=2.9.5:83)
    at HTMLLIElement.dispatch (jquery-1.9.1.js?version=2.9.5:3074)
```

wimpers:
@jeroenmaelbrancke  after a retry it worked, right?

jeroenmaelbrancke:
Refreshing the page fixed the problem.

JeffreyDevloo:
Fixed by 
- #1821 
- https://github.com/openvstorage/framework-iscsi-plugin/pull/34
- https://github.com/openvstorage/framework-extensions/pull/63
- https://github.com/openvstorage/framework-alba-plugin/pull/534

Packaged in:
 - openvstorage_2.12.0-dev.1518011126.4cee8bc-1_amd64.deb
 - openvstorage-backend_1.12.0-dev.1518011188.e274a3d-1_amd64.deb
 - openvstorage-extensions_0.4.0-dev.1518010323.90e5a68-1_amd64.deb
 - openvstorage-iscsi-plugin_0.4.0-dev.1518011220.53302ad-1_amd64.deb

JeffreyDevloo:
Released in https://github.com/openvstorage/framework/releases/tag/2.12.0
Packaged in openvstorage-core_2.12.0-1_amd64.deb

Issue #1658, opened 2017-06-06
Change storage router before the ips are loading result in wrong ip (create vpool)


jeroenmaelbrancke:
When changing the storage router while the ips are still loading result in a wrong list of ips.

Before the ips are loaded:
![wrong_ip](https://cloud.githubusercontent.com/assets/2900612/26822563/c430bc5c-4aaa-11e7-9479-ad18efb83464.png)

Waiting until the ips are loaded:

![correct_ip](https://cloud.githubusercontent.com/assets/2900612/26822567/c7021e58-4aaa-11e7-83da-1507d44714aa.png)

JeffreyDevloo:
Taking this up with my edit-vpool changes. Have to change certain aspects about the add, might as well take this up.

JeffreyDevloo:
Fixed by 
- #1821 
- https://github.com/openvstorage/framework-iscsi-plugin/pull/34
- https://github.com/openvstorage/framework-extensions/pull/63
- https://github.com/openvstorage/framework-alba-plugin/pull/534

Packaged in:
 - openvstorage_2.12.0-dev.1518011126.4cee8bc-1_amd64.deb
 - openvstorage-backend_1.12.0-dev.1518011188.e274a3d-1_amd64.deb
 - openvstorage-extensions_0.4.0-dev.1518010323.90e5a68-1_amd64.deb
 - openvstorage-iscsi-plugin_0.4.0-dev.1518011220.53302ad-1_amd64.deb

JeffreyDevloo:
Released in https://github.com/openvstorage/framework/releases/tag/2.12.0
Packaged in openvstorage-core_2.12.0-1_amd64.deb

PR #1821, opened 2017-10-09
Reconfigure vpool

JeffreyDevloo:


JeffreyDevloo:
This pull request only refactors the whole add vpool and has multiple fixes for other front-end stuff too:
List of items: 
 - https://github.com/openvstorage/framework/issues/1678
 - https://github.com/openvstorage/framework/issues/1669
 - https://github.com/openvstorage/framework/issues/1658

Items to continue (foundation is built, but not yet finished)
 - https://github.com/openvstorage/framework/issues/1687: struggled with syncing the models, has to be implemented on the api side and validation should be put in
 - https://github.com/openvstorage/framework/issues/1656: front-end wise everything is built, has to be wrapped up on the api side

General front-end rework: Multiple containers have been reworked to map to the api for easier updating (should be applied to all containers)

The actual reconfiguring is currently on hold as the change list is becoming too big to review

ovs/dal/hybrids/storagedriver.py
```diff
@@ -122,55 +124,33 @@ def _vpool_backend_info(self):
...
-                'cache_quota_fc': cache_quota_fc,
-                'cache_quota_bc': cache_quota_bc,
-                'backend_info': backend_info,
-                'connection_info': connection_info,
-                'block_cache_read': block_cache_read,
-                'block_cache_write': block_cache_write,
-                'block_cache_backend_info': block_cache_backend_info,
-                'block_cache_connection_info': block_cache_connection_info,
-                'global_write_buffer': global_write_buffer}
+        from ovs.lib.storagedriver import StorageDriverController
```
kvanhijf: Since importing controller logic in the DAL isn't a good thing to do, i'd opt for adding another dynamic property to the StorageDriver hybrid -> `self._global_write_buffer` or something like that

PS: I wouldn't use a normal property because WRITE roles can be added
PS2: Perhaps the global write buffer should not be returned by this dynamic anymore if we create a separate dynamic for it

ovs/dal/hybrids/storagedriver.py
```diff
@@ -122,55 +124,33 @@ def _vpool_backend_info(self):
...
+                                                               'write': False,
+                                                               'quota': None,
+                                                               'backend_info': None}}}
+        if 'caching_info' not in self.vpool.metadata:
+            self._logger.critical('Metadata structure has not been updated yet')
+            return vpool_backend_info
+        if self.storagerouter_guid not in self.vpool.metadata['caching_info']:
+            # No caching configured
+            return vpool_backend_info
+        for cache_type, cache_data in vpool_backend_info['caching_info'].iteritems():
```
kvanhijf, line 146: A lot of potential KeyErrors here, thats why i'd put this in a try - except because API calls querying for this information will fail and result in the GUI becoming unresponsive

ovs/dal/hybrids/storagedriver.py
```diff
@@ -122,55 +124,33 @@ def _vpool_backend_info(self):
...
+            caching_info = self.vpool.metadata['caching_info'][self.storagerouter_guid][cache_type]
+            # Update the cache data matching the keys currently specified in cache_data
+            cache_data.update((k, caching_info[k]) for k in cache_data.viewkeys() & caching_info.viewkeys())
```
kvanhijf, line 149: Since a KeyError will occur if there are mismatches between keys in `cache_data` and `caching_info`, i don't think its a good thing to take a combined set and loop over them

ovs/dal/hybrids/storagedriver.py
```diff
@@ -122,55 +124,33 @@ def _vpool_backend_info(self):
...
-                'block_cache_backend_info': block_cache_backend_info,
-                'block_cache_connection_info': block_cache_connection_info,
-                'global_write_buffer': global_write_buffer}
+        from ovs.lib.storagedriver import StorageDriverController
+        global_write_buffer = StorageDriverController.calculate_global_write_buffer(self.guid)
+        vpool_backend_info = {'backend': copy.deepcopy(self.vpool.metadata['backend']),
+                              'caching_info': {'fragment_cache': {'read': False,
+                                                                  'write': False,
+                                                                  'quota': None,
+                                                                  'backend_info': None},  # Will contain connection info if it wouldn't be None
```
kvanhijf: When / where is the connection info passed in?

ovs/dal/hybrids/storagedriver.py
```diff
@@ -212,3 +192,25 @@ def _cluster_node_config(self):
...
+        proxy_info = {'red': 0,
+                      'orange': 0,
+                      'green': 0}
+        summary = {'proxies': proxy_info}
+        for alba_proxy in self.alba_proxies:
+            service_status = service_manager.get_service_status(alba_proxy.service.name, client)
+            if service_status == 'active':
+                proxy_info['green'] += 1
+            elif service_status == 'inactive':
+                proxy_info['inactive'] += 1
```
kvanhijf: `inactive` is not a valid key
also add an `else` stack to increase the `red` amount

ovs/dal/hybrids/storagedriver.py
```diff
@@ -212,3 +192,25 @@ def _cluster_node_config(self):
...
+    def _local_summary(self):
+        """
+        Returns a summary of the proxies of this storagedriver
+        :return: summary of the proxies
+        :rtype: dict
+        """
+        from ovs.extensions.generic.sshclient import SSHClient
+        from ovs.extensions.services.servicefactory import ServiceFactory
+        service_manager = ServiceFactory.get_manager()
+        client = SSHClient(self.storagerouter)
```
kvanhijf: This can throw UnableToConnectException and NotAuthenticatedException, so put it in a try - except plz

ovs/dal/hybrids/storagedriver.py
```diff
@@ -212,3 +192,25 @@ def _cluster_node_config(self):
                                                              self.storage_ip,
                                                              self.ports['edge']),
                 'node_distance_map': distance_map}
+
+    def _local_summary(self):
```
kvanhijf: Name doesn't describe the fact it returns proxy information

ovs/dal/hybrids/storagedriver.py
```diff
@@ -212,3 +192,25 @@ def _cluster_node_config(self):
...
+        """
+        Returns a summary of the proxies of this storagedriver
+        :return: summary of the proxies
+        :rtype: dict
+        """
+        from ovs.extensions.generic.sshclient import SSHClient
```
kvanhijf: Add these imports on top
None of the imports should do DAL imports, so no reason for circular imports

ovs/lib/migration.py
```diff
@@ -245,6 +245,72 @@ def migrate():
...
 
+        #####################################
+        # Update the vPool metadata structure
+        def _update_metadata_structure(metadata):
+            metadata = copy.deepcopy(metadata)
+            cache_structure = {'read': False,
+                               'write': False,
+                               'is_backend': False,
+                               'quota': None,
+                               'backend_info': {'name': None,  # Will be filled in when isBackend is true
```
kvanhijf: Should be `is_backend` in comment

ovs/lib/migration.py
```diff
@@ -245,6 +245,72 @@ def migrate():
                 vpool.metadata_store_bits = bits
                 vpool.save()
 
+        #####################################
```
kvanhijf, line 255: Should actually entirely be moved to the out of band migration section

ovs/lib/migration.py
```diff
@@ -245,6 +245,72 @@ def migrate():
...
+                                new_cache_structure['backend_info']['arakoon_config'] = backend_data['arakoon_config']
+                                new_cache_structure['backend_info'].update(backend_data['backend_info'])
+                                new_cache_structure['backend_info']['connection_info'].update(backend_data['connection_info'])
+                            else:
+                                new_cache_structure[new_structure_key] = current_cache_data.get(old_structure_key)
+            return metadata
+
+        vpools = VPoolList.get_vpools()
+        for vpool in vpools:
+            new_metadata = _update_metadata_structure(vpool.metadata)
```
kvanhijf: Can potentially throw KeyErrors and perhaps other errors too, so i'd add this in a try except

ovs/lib/monitoring.py
```diff
@@ -71,14 +71,14 @@ def verify_vdisk_cache_quota():
...
-
-            for cache_type in [['cache_quota_fc', 'backend_info', 'connection_info', 'fragment'],
-                               ['cache_quota_bc', 'block_cache_backend_info', 'block_cache_connection_info', 'block']]:
-                cache_quota = storagedriver.vpool_backend_info[cache_type[0]]
-                backend_info = storagedriver.vpool_backend_info[cache_type[1]]
-                connection_info = storagedriver.vpool_backend_info[cache_type[2]]
+            for cache_type, cache_type_data in storagedriver.vpool_backend_info['caching_info'].iteritems():
+                cache_quota = cache_type_data['quota']
+                backend_info = cache_type_data['backend_info']
+                connection_info = cache_type_data.get('backend_info', {}).get('connection_info', None)
```
kvanhijf, line 82: `backend_info` is always filled out in the new structure, being `None` when there is no backend involved, resulting in an error when doing the `get`. Instead the `vpool_backend_info` should always return a dict for `backend_info` imo

ovs/lib/storagedriver.py
```diff
@@ -17,27 +17,36 @@
...
 from ovs.dal.lists.storagerouterlist import StorageRouterList
 from ovs.dal.lists.vdisklist import VDiskList
 from ovs.dal.lists.vpoollist import VPoolList
 from ovs.extensions.db.arakooninstaller import ArakoonClusterConfig, ArakoonInstaller
 from ovs.extensions.generic.configuration import Configuration
 from ovs.extensions.generic.logger import Logger
 from ovs_extensions.generic.remote import remote
+from ovs.extensions.generic.volatilemutex import volatile_mutex
 from ovs.extensions.generic.sshclient import NotAuthenticatedException, SSHClient, UnableToConnectException
+from ovs.lib.helpers.toolbox import Toolbox
```
kvanhijf: Get your sortings straight

ovs/lib/storagedriver.py
```diff
@@ -399,6 +408,200 @@ def _configure_arakoon_to_volumedriver(cluster_name):
                                                                           dls_arakoon_cluster_nodes=arakoon_nodes)
                     storagedriver_config.save()
 
+    @staticmethod
+    def create_new_storagedriver(vpool_guid, storagerouter_guid, storage_ip, amount_of_proxies):
```
kvanhijf: Tautology detected + wouldn't it be niftier to create some sort of StorageDriverConfigurator / StorageDriverInstaller class which can be initiated with a vpool guid, sr guid and such and then add proxies to it, configure stuff, ...
Then we don't need to pass all the information through dicts and many args and kwargs

ovs/lib/storagedriver.py
```diff
@@ -399,6 +408,200 @@ def _configure_arakoon_to_volumedriver(cluster_name):
...
+        """
+        # * Information about backoff_gap and trigger_gap (Reason for 'smallest_write_partition' introduction)
+        # * Once the free space on a mount point is < trigger_gap (default 1GiB), it will be cleaned up and the cleaner attempts to
+        # * make sure that <backoff_gap> free space is available => backoff_gap must be <= size of the partition
+        # * Both backoff_gap and trigger_gap apply to each mount point individually, but cannot be configured on a per mount point base
+        # Assign WRITE / Fragment cache
+        from ovs.lib.storagerouter import StorageRouterController  # Avoid circular reference
+        storagedriver = StorageDriver(storagedriver_guid)
+        storagerouter = storagedriver.storagerouter
+        if partition_info is None:
```
kvanhijf: There's only 1 usage of this function in which you always pass the partition information

kvanhijf (dismissed):

JeffreyDevloo:
Still an issue with local caching... The wizard is blocking

ovs/dal/hybrids/storagedriver.py
```diff
@@ -24,8 +24,11 @@
 from ovs.dal.structures import Property, Relation, Dynamic
 from ovs.dal.hybrids.vdisk import VDisk
 from ovs.dal.hybrids.vpool import VPool
+from ovs.dal.hybrids.diskpartition import DiskPartition
```
sjovdnbos, line 27: Alphabetical order needed here or not?
If so: 
1. diskpartition
2. storagerouter
3. vdisk
4. vpool

sjovdnbos (approved):

Issue #1822, opened 2017-10-10
Scrubber deletes parent directory


jeroenmaelbrancke:
The scrubber clean up the scrubber directory even if other scrubber are running using this parent directory:

for example:
root@NY1SRV1000:~# ls /mnt/hdd2/scrub_work_ss-ny1-01_NY1SRV1000/
1d73f7e6-52af-40e1-903d-26b0f89e12af  56873b39-3a4e-4847-b7ba-e4c9d3fee23f  5ee9e5ca-8d3e-4bf4-9e1a-8f2cb0b9250c

if scrubber for vdisk 1d73f7e6-52af-40e1-903d-26b0f89e12af finished he will remove the complete directory scrub_work_ss-ny1-01_NY1SRV1000. The other scrubbers will crash as they cannot write to the scrubber directory.

for example:
2017-10-10 06:55:48 933427 -0400 - NY1SRV1000 - 4492/0x00007f85bea99900 - scrubber/FileUtils - 000000000000008f - error - create_temp_file: Failed to create temp file /mnt/hdd2/scrub_work_data-ny2-01_NY1SRV1000/31eb5c23-3afc-4b6c-a6ac-184
fbe1622ed/00_00000cc7_00WlBbdV: No such file or directory

kvanhijf:
Included in release 2.10.2

PR #1826, opened 2017-10-11
#1822 - Fix to run multiple single vdisk scrub jobs simultaneously

kvanhijf:


JeffreyDevloo (approved):
Before packaging: promote the current develop which contains other bugfixes

PR #1858, opened 2017-11-07
Change token lifetime and make them configurable

JeffreyDevloo:


webapps/api/oauth2/tokenview.py
```diff
@@ -31,6 +31,7 @@
 from ovs.dal.lists.userlist import UserList
 from ovs.dal.lists.rolelist import RoleList
 from ovs.dal.hybrids.client import Client
+from ovs.extensions.generic.configuration import Configuration
```
kvanhijf: nice try

webapps/api/oauth2/tokenview.py
```diff
@@ -47,7 +52,12 @@ def post(self, request, *args, **kwargs):
         """
         Handles token post
         """
-        logger = Logger('api')
+        expiration = Configuration.get('ovs/framework/api/oauth|expiration', default={})
+        if not isinstance(expiration, dict):
+            self.logger.exception('The oauth expiration configuration is not an object. Falling back to defaults')
+            expiration = {}
+        expiration_password = expiration.get('password', self.EXPIRATION_PASSWORD)
```
kvanhijf: Why a dict in a dict?
now the value must look like
{'expiration': {'password': 3600, 'client': 3600}}

Why not just use this structure:
{'password_expiration': 3600, 'client_expiration': 3600}

And while writing this i was thinking whether this is a security risk?  Every user is able to modify the timeouts of tokens?  Do we want this?

kvanhijf (changes requested):

webapps/api/oauth2/tokenview.py
JeffreyDevloo: Does ovs config get/set work for any user except for root and ovs?
And with that logic, a user can wreck havoc in other parts of our code too

Security risk around tokens is that users could use the api longer than we really wanted them (given that someone was able to edit this token point instead of rewriting our code).
However we store the tokens within our DAL and this means we have a way of revoking them when shit hits the fan, so there is a minor risk (as there is now) but it is counterable

The dict in a dict part, I found it more neat to categorize by type instead of prepending

webapps/api/oauth2/tokenview.py
```diff
@@ -47,7 +49,8 @@ def post(self, request, *args, **kwargs):
         """
         Handles token post
         """
-        logger = Logger('api')
+        expiration_user = Configuration.get('ovs/framework/api/oauth|expiration_user', default=OAuth2Toolbox.EXPIRATION_USER)
+        expiration_client = Configuration.get('ovs/framework/api/oauth|expiration_user', default=OAuth2Toolbox.EXPIRATION_CLIENT)
```
kvanhijf: incorrect key should be `expiration_client`

kvanhijf (changes requested):

webapps/api/oauth2/tokenview.py
JeffreyDevloo: I will change the time out in the unit tests as this should have been picked up by them :p

webapps/api/tests/test_authentication.py
```diff
@@ -354,7 +370,20 @@ def test_client_credentials(self):
         self.assertIn('access_token', response_content)
         result = {'access_token': response_content['access_token'],
                   'token_type': 'bearer',
-                  'expires_in': 3600}
+                  'expires_in': OAuth2Toolbox.EXPIRATION_CLIENT}
+        self.assertDictEqual(response_content, result)
+
+        # Test config expiration
+        week_expiration = 60 * 60 * 24 * 3
```
kvanhijf: I hope these weeks of 3 days contain 2 weekend days

webapps/api/tests/test_authentication.py
```diff
@@ -354,7 +370,20 @@ def test_client_credentials(self):
...
+        Configuration.set('ovs/framework/api/oauth|expiration_client', week_expiration)
+        request = self.factory.post('/', data=data, HTTP_X_REAL_IP='127.0.0.6', HTTP_AUTHORIZATION=header)
+        response = OAuth2TokenView.as_view()(request)
+        self.assertEqual(response.status_code, 200)
+        response_content = json.loads(response.content)
+        self.assertIn('access_token', response_content)
+        result = {'access_token': response_content['access_token'],
+                  'token_type': 'bearer',
+                  'expires_in': OAuth2Toolbox.EXPIRATION_CLIENT}
```
kvanhijf: Shouldn't the `expires_in` be the `week_expiration` here?

kvanhijf (changes requested):

kvanhijf:
Ticket has been put on roadmap. Should we merge this workaround in the code?

JeffreyDevloo:
It would give ops the flexibility to configure it to their liking but this does not really fix anything. So it is your call

wimpers:
please create a ticket for the code you want to merge

JeffreyDevloo:
Ticket: https://github.com/openvstorage/framework/issues/1860

kvanhijf (approved):

PR #1880, opened 2017-11-27
List filter

JeffreyDevloo:


ovs/dal/tests/test_basic.py
```diff
@@ -244,89 +245,178 @@ def test_queries(self):
...
         dlist = DataList(TestDisk, {'type': DataList.where_operator.AND,
                                     'items': [('size', DataList.operator.EQUALS, 1)]})
-        self.assertEqual(len(dlist), 1, 'list should contain 1')
-        dlist = DataList(TestDisk, {'type': DataList.where_operator.AND,
-                                    'items': [('size', DataList.operator.EQUALS, 1)]})
-        found_object = dlist[0]
-        self.assertEqual(found_object.name, 'test_1', 'list should contain correct machine')
+        expected_disks = [TestDisk(disk_guids[1])]
+        self.assertEqual(len(dlist), len(expected_disks), 'List should contain 1')
+        self.assertItemsEqual(list(dlist), expected_disks, 'List should contain disk 1')
```
kvanhijf: In the new test you wrote, you never converted the DataList to a regular python list?  So is this necessary?

ovs/dal/tests/test_basic.py
```diff
@@ -521,6 +611,156 @@ def test_listcache(self):
...
+        self.assertItemsEqual(all_list0_guid, expected_items, 'List should return machine_0-9')
+        self.assertEqual(all_list0_guid.from_cache, False, 'List should not come from cache')
+        self.assertNotEqual(all_list0._key, all_list0_guid._key, 'Keys should differ')
+
+        all_list1_guid = DataList(TestMachine, guids=machine_guids)
+        expected_items = [TestMachine(guid) for guid in machine_guids]
+        self.assertEqual(len(all_list1_guid), len(expected_items), 'List should return 10')
+        self.assertItemsEqual(all_list1_guid, expected_items, 'List should return machine_0-9')
+        self.assertEqual(all_list1_guid.from_cache, True, 'List should come from cache')
+
```
kvanhijf: Check on the `key` here? `all_list1_guid._key == all_list0_guid._key` right?

ovs/dal/tests/test_basic.py
```diff
@@ -521,6 +611,156 @@ def test_listcache(self):
...
+        key = uuid.uuid4()
+        ############
+        # With key #
+        ############
+        all_list0 = DataList(TestMachine, key=key)
+        expected_items = [TestMachine(guid) for guid in machine_guids]
+        self.assertEqual(len(all_list0), len(expected_items), 'List should return 10')
+        self.assertItemsEqual(all_list0, expected_items, 'List should return machine_0-9')
+        self.assertEqual(all_list0.from_cache, False, 'List should not come from cache')
+
```
kvanhijf, line 756: Verify that the key of the returned DataList is the key passed?  (times 3 below)

ovs/dal/datalist.py
```diff
@@ -103,16 +116,132 @@ def guids(self):
...
+    def _validate_guids(guids):
+        """
+        Validates if the supplied guids are valid
+        :param guids: Guids to check for
+        :type guids: list[basestring]
+        :return: None
+        :rtype: NoneType
+        :raises: ValueError if the guids are not valid
+        """
+        if guids is not None and not isinstance(guids, list) and (len(guids) > 0 and isinstance(guids[0], basestring)):
```
kvanhijf: You'll never reach the final statement (`len(guids) > 0 and isinstance(guids[0], basestring)`) in case you have a list
Which means you'll never check the 1st element in the list

and why not check all items in the list whether its a string?

ovs/dal/datalist.py
```diff
@@ -103,16 +116,132 @@ def guids(self):
...
+    def _validate_query(query):
+        """
+        Validates if a query is of the format we'd expect
+        :param query: Query to perform
+        :type query: dict
+        :return: None
+        :rtype: NoneType
+        :raises: ValueError if the query is not valid
+        """
+        if query is not None and not isinstance(query, dict) and not all(k in query for k in ("type", "items")):
```
kvanhijf: You'll never reach the final statement (`not all(k in query for k in ("type", "items"))`) in case you have a dict
Which means you'll never check the keys in the dict

Perhaps a small unitttest to verify this?

ovs/dal/datalist.py
```diff
@@ -90,9 +98,14 @@ def __init__(self, object_type, query, key=None):
         self._guids = None
         self._executed = False
         self._shallow_sort = True
-
         self.from_cache = None
         self.from_index = 'none'
+        self._provided_guids = guids
```
kvanhijf: I preferred the older notation where public/private properties were separated

ovs/dal/datalist.py
```diff
@@ -103,16 +116,132 @@ def guids(self):
...
+        :type key: str
+        :return: None
+        :rtype: NoneType
+        """
+        if key is not None:
+            self._key = '{0}_{1}'.format(DataList.NAMESPACE, key)
+            self._provided_key = True
+            # Unsure whether or not the same query would apply
+            self._volatile.delete(self._key)
+            return
```
kvanhijf: Obsolete return statement

ovs/dal/datalist.py
```diff
@@ -103,16 +116,132 @@ def guids(self):
...
+        elif self._provided_key is False:
+            identifier = copy.deepcopy(self._query)
+            identifier['object'] = self._object_type.__name__
+            identifier['guids'] = 'None' if self._provided_guids is None else ','.join(self._provided_guids)
```
kvanhijf: I'd order the guids, because a different order of the same guids should return the same result

ovs/dal/datalist.py
```diff
@@ -103,16 +116,132 @@ def guids(self):
...
+        :return: None
+        :rtype: NoneType
+        """
+        if key is not None:
+            self._key = '{0}_{1}'.format(DataList.NAMESPACE, key)
+            self._provided_key = True
+            # Unsure whether or not the same query would apply
+            self._volatile.delete(self._key)
+            return
+        elif self._provided_key is False:
```
kvanhijf: As far as i can tell, its impossible to reset the key
If you create a DataList without a key --> `self._provided_key` will be `False`
If you then use `set_key` on that DataList --> `self._provided_key` will be `True`
But there is no way to reset the provided key right?  Not without tampering with the private properties

ovs/dal/datalist.py
```diff
@@ -103,16 +116,132 @@ def guids(self):
...
+            self._key = '{0}_{1}'.format(DataList.NAMESPACE, hashlib.sha256(json.dumps(identifier)).hexdigest())
+
+    def _reset_list(self):
+        """
+        Resets everything about the DataList
+        :return: None
+        :rtype: NoneType
+        """
+        # Force query to rerun
+        self._executed = False
```
kvanhijf, line 149: Wouldn't it be more clean to re-run the `__init__` here instead?

ovs/dal/datalist.py
```diff
@@ -103,16 +116,132 @@ def guids(self):
...
+        * <filter>: A tuple defining a single expression:
+                    (<field>, DataList.operator.XYZ, <value> [, <ignore_case>])
+                    The field is any property you would also find on the given object. In case of
+                    properties, you can dot as far as you like.
+        * <items>: A list of one or more <query> or <filter> items. This means the query structure is recursive and
+                   complex queries are possible
+        :return: None
+        :rtype: NoneType
+        """
+        if query is None:
```
kvanhijf, line 178: Don't know for sure whether `__init__` could be called here too. Since `reset_list` is called upon here
If possible, you don't need to validate the query, nor delete the volatile key

If not possible, you need to check whether the newly provided query is different from previously stored query
If not, no action is needed

ovs/dal/datalist.py
```diff
@@ -103,16 +116,132 @@ def guids(self):
...
+
+    def set_guids(self, guids):
+        """
+        Sets up a list of guids to apply the query too
+        :param guids: List of guids to query on or None in case you wish to query all items again
+        :type guids: list[basestring] or NoneType
+        :return: None
+        :rtype: NoneType
+        """
+        if guids is not None:
```
kvanhijf: Obsolete if statement

ovs/dal/datalist.py
```diff
@@ -103,16 +116,132 @@ def guids(self):
...
+            raise ValueError('Specified guids should be a list of guids or None')
+
     #######################
     # Query functionality #
     #######################
 
     def _get_keys_from_index(self, indexed_properties, items, where_operator):
         """
-        Builds a generator that only yields data filtered by possible indexes where possible.
+        Builds a set of keys that were retrieved from the indexes.
```
kvanhijf, line 241: 1 index, 2 indices

webapps/api/backend/decorators.py
```diff
@@ -257,21 +259,32 @@ def wrap(f):
...
+            query = request.QUERY_PARAMS.get('query')
             if sort is None and default_sort is not None:
                 sort = default_sort
             sort = None if sort is None else [s for s in reversed(sort.split(','))]
             page = request.QUERY_PARAMS.get('page')
-            page = int(page) if page is not None and page.isdigit() else None
+            page = int(page) if page is not None and (isinstance(page, int) or page.isdigit()) else None
             page_size = request.QUERY_PARAMS.get('page_size')
-            page_size = int(page_size) if page_size is not None and page_size.isdigit() else None
-            page_size = page_size if page_size in [10, 25, 50, 100] else 10
```
kvanhijf, line 274: Why did you remove these default sizes?
Our GUI only supports these values, so shouldn't we keep this?

webapps/api/backend/decorators.py
```diff
@@ -289,11 +302,25 @@ def new_function(*args, **kwargs):
             guid_list = isinstance(data_list, list) and len(data_list) > 0 and isinstance(data_list[0], basestring)
             timings['fetch'] = [time.time() - start, 'Fetching data']
 
+            # Filtering data
```
kvanhijf: Why not apply the query right away in the `Fetch data` step?

webapps/api/backend/decorators.py
```diff
@@ -289,11 +302,25 @@ def new_function(*args, **kwargs):
...
+                data_list = DataList(object_type, query=query, guids=guids)
+                # Trigger the query
+                _ = data_list.guids
+                timings['querying'] = [time.time() - start, 'Querying data']
+
             # 4. Sorting
             if sort is not None:
                 start = time.time()
                 if guid_list is True:
-                    data_list = [object_type(guid) for guid in data_list]
```
kvanhijf: Instantiating a new DataList results in the `._executed` to be `False`
The old notation resulted in an actual list with fully serialized objects, does the new notation have the same result?

kvanhijf (changes requested):

ovs/dal/tests/test_basic.py
JeffreyDevloo: No, it is not a copy of the old list but a pointer. So the old list got changed too

webapps/api/backend/decorators.py
JeffreyDevloo: Our GUI could support other sizes too. Our dropdown is only limited to those.
No need to keep the defaults as it is the callee's problem when their pagination is not returning their number of items

Also it is a bit nasty to not reject but instead go by the default of 10

ovs/dal/datalist.py
JeffreyDevloo: Not really required to check the query difference as the hash would be same and thus the caching value would be the same (covered in the unittest)
Deleting the volatile key is necessary when a key was specified as this key should always be used for this DataList instance. When not clearing the cache, changing the query or guids would return results from the previous _execute_query which is not what we'd want

ovs/dal/datalist.py
JeffreyDevloo: I had no idea you could tbh and I'm also against it. An __init__ is an initialization function, even though Python does not care and could recycle your instance. For other coders who know they can only initialize something once, this might lead to confusion (and it's in general a bad choice)

ovs/dal/datalist.py
JeffreyDevloo: http://grammarist.com/usage/indexes-indices/

ovs/dal/tests/test_basic.py
JeffreyDevloo: The key is never identical to the guid generated because of the namespace that is prepended.
Also, will take this up on the init test

webapps/api/backend/decorators.py
JeffreyDevloo: To make it pretty
The sorting could also be applied in that step really

webapps/api/backend/decorators.py
JeffreyDevloo: Yes, although serialization is still done when 'contents' is requested.
Now instead of manually creating a serialized list, everything is offloaded to DataList
The test_decorators tests this behaviour

ovs/dal/tests/test_basic.py
JeffreyDevloo: No, requires an iterable (which DataList is)

ovs/dal/tests/test_basic.py
```diff
@@ -591,7 +605,10 @@ def test_listcache(self):
                                   query={'type': DataList.where_operator.AND,
                                          'items': [('machine.name', DataList.operator.EQUALS, 'machine')]})
             list_cache._execute_query()
-            self.assertTrue(list_cache.from_cache, 'List should be loaded from cache (mode: {0})'.format(key))
+            if key is None:
+                self.assertTrue(list_cache.from_cache, 'List should be loaded from cache (mode: {0})'.format(key))
+            else:
+                self.assertFalse(list_cache.from_cache, 'List should be loaded from cache (mode: {0})'.format(key))
```
kvanhijf: Comment should be `List should not be loaded from ....`

ovs/dal/tests/test_basic.py
```diff
@@ -591,7 +605,10 @@ def test_listcache(self):
                                   query={'type': DataList.where_operator.AND,
                                          'items': [('machine.name', DataList.operator.EQUALS, 'machine')]})
             list_cache._execute_query()
-            self.assertTrue(list_cache.from_cache, 'List should be loaded from cache (mode: {0})'.format(key))
+            if key is None:
+                self.assertTrue(list_cache.from_cache, 'List should be loaded from cache (mode: {0})'.format(key))
```
kvanhijf, line 633: This message will contain `mode: None`
Is this expected?

kvanhijf (changes requested):

ovs/dal/tests/test_basic.py
JeffreyDevloo: Yes, it was always like that

kvanhijf (approved):

PR #1912, opened 2017-12-27
docstring updated

sjovdnbos:


ovs/lib/update.py
```diff
@@ -555,6 +555,8 @@ def execute_update(components):
         """
         Update the specified components on all StorageRouters
         This is called upon by 'at'
+        :param components: list of the components to update
+        :type components: []
```
domsj: `:type components: list` ?
(at least according to a quick grep around the codebase that seems to be the way it's done elsewhere)

domsj (approved):

PR #1913, opened 2017-12-27
Incorrect execStopPost in ovs-voldrv.service

sjovdnbos:
Previous implementation led to dangerous behaviour where all volumedrivers would have been killed where only the voldrv for one vpool is intended to be removed


domsj (approved):

config/templates/systemd/ovs-volumedriver.service
```diff
@@ -9,7 +9,7 @@ Environment=METADATASTORE_BITS=<METADATASTORE_BITS>
...
 ExecStart=/usr/bin/volumedriver_fs.sh -f --config <CONFIG_PATH> --lock-file <RUN_FILE_DIR>/storagedriver_<VPOOL_NAME>.lock --logrotation --mountpoint <VPOOL_MOUNTPOINT> --logsink <LOG_SINK> -o big_writes -o sync_read -o allow_other -o use_ino -o default_permissions -o uid=<OVS_UID> -o gid=<OVS_GID> -o umask=0002
 ExecStop=/bin/bash -c "if mount | grep <VPOOL_MOUNTPOINT>; then fusermount -u <VPOOL_MOUNTPOINT>; fi "
-ExecStopPost=/bin/bash -c "for i in `seq 0 9`; do if mount | grep <VPOOL_MOUNTPOINT>; then sleep 10s; else break; fi; done; if mount | grep <VPOOL_MOUNTPOINT>; then umount -l <VPOOL_MOUNTPOINT>; ps xa |grep volumedriver_fs |grep -v grep |awk '{print $1}' |xargs kill; fi; rm <RUN_FILE_DIR>/storagedriver_<VPOOL_NAME>.lock"
+ExecStopPost=/bin/bash -c "for i in `seq 0 9`; do if mount | grep <VPOOL_MOUNTPOINT>; then sleep 10s; else break; fi; done; if mount | grep <VPOOL_MOUNTPOINT>; then umount -l <VPOOL_MOUNTPOINT>; pkill -f -- 'volumedriver_fs .*--mountpoint <VPOOL_MOUNTPOINT>'; fi; rm <RUN_FILE_DIR>/storagedriver_<VPOOL_NAME>.lock"
```
dejonghb, line 12: the "grep <VPOOL_MOUNT>" is too wide: eg. /mnt/vp1 & /mnt/vp12 --> grepping for /mnt/vp1 will return both which results in the for loop being run 10 times = 100 seconds (same holds if /mntvp1 isn't mounted any more but /mnt/vp12 is still active)

grepping for lines with ' on <VPOOL_MOUNTPOINT> type ' (including space at start and end)  after the path should prevent this 
```
mount | grep 'on <VPOOL_MOUNTPOINT> type '
```

alternative: /proc/mounts is a pseudofile with all mounts (and as long as there are no space in the mountpoint ;-) )
```
grep ' <VPOOL_MOUNTPOINT> ' /proc/mounts
```

sjovdnbos:
was continued in pr #1914
package info: openvstorage-2.11.1-dev.1514385624.eeb011e

Issue #1835, opened 2017-10-16
API should allow for generic calls


kvanhijf:
Our current implementation does not allow to add generic calls which can be used for an entire cluster for example. Eg: client.get('get_cluster_status')

This is currently impossible, because we require a DAL object to be passed in the API, like this
client.get('storagerouters/' + storagerouter.guid + '/get_cluster_status')

This is a very sub-awesome implementation for an API

JeffreyDevloo:
Taking this up as I need for another ticket.

JeffreyDevloo:
This would require a update from the Django-rest-framework. 
http://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-methods-for-routing or we could test https://github.com/chibisov/drf-extensions which does the same for older Django versions.
I'd opt for an update

wimpers:
@JeffreyDevloo  what is the status of this ticket? Did you  pick it up? Why is this urgent?

JeffreyDevloo:
This is still pending, to be picked up. I did not require it so far (as the reconfigure storagedriver was put on hold for reviewing)

wimpers:
@JeffreyDevloo so you need it for finishing the reconfigure storagedriver ticket?

JeffreyDevloo:
Not really, but it would mean that an api call to retrieve cluster-wide information is bound to a parameterized api.
Example: get_all_statistics, ideally it would be placed on vpool so /api/vpools/get_all_statistics
However, the current implementation does not allow this and one would need to provide an id: /api/vpool/<vpool_guid>/get_all_statistics
Placing it on a call with a <vpool_guid> makes it look like that the call is only for that instance, while it would actually retrieve all statistics from every vpool

wimpers:
Removed Urgent label

JeffreyDevloo:
Tackling as the iscsi-ha changes require even more generic calls

JeffreyDevloo:
Fixed by https://github.com/openvstorage/framework/pull/2039
Packaged in openvstorage-core_2.12.0-dev.1525420669.4a3f34e-1_amd64.deb

JeffreyDevloo:
Released in https://github.com/openvstorage/framework/releases/tag/2.12.0
Packaged in openvstorage-core_2.12.0-1_amd64.deb

PR #1917, opened 2018-01-02
Move multiple vdisks

sjovdnbos:
Moval of multiple vdisks to a single storagerouter

### DIFF
--- a/webapps/api/backend/views/vdisks.py
+++ b/webapps/api/backend/views/vdisks.py
@@ -222,27 +222,6 @@ class VDiskViewSet(viewsets.ViewSet):
     @required_roles(['read', 'write'])
     @return_task()
     @load(VDisk)
-    def move_multiple(self, vdisk, target_storagerouter_guid, force=False):
-        """
-        Moves a vDisk
-        :param vdisk: Guids of the virtual disk to move
-        :type vdisk: list
-        :param target_storagerouter_guid: Guid of the StorageRouter to move the vDisks to
-        :type target_storagerouter_guid: str
-        :param force: Indicate whether to force the migration (forcing the migration might cause data loss)
-        :type force: bool
-        :return: Asynchronous result of a CeleryTask
-        :rtype: celery.result.AsyncResult
-        """
-        return VDiskController.move.delay(vdisk_guid=vdisk.guid,
-                                          target_storagerouter_guid=target_storagerouter_guid,
-                                          force=force)
-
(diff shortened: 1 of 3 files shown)


webapps/api/backend/views/vdisks.py
```diff
@@ -217,6 +217,27 @@ def move(self, vdisk, target_storagerouter_guid, force=False):
                                           target_storagerouter_guid=target_storagerouter_guid,
                                           force=force)
 
+    @action()
+    @log()
+    @required_roles(['read', 'write'])
+    @return_task()
+    @load(VDisk)
+    def move_multiple(self, vdisk, target_storagerouter_guid, force=False):
```
JeffreyDevloo: This indicates the API would be on /api/vdisks/<vdisk_guid>/move_multiple which is not what I'd expect.
Due to our Django not supporting generic calls atm (https://github.com/openvstorage/framework/issues/1835), I'd opt to put it on the vpool level

JeffreyDevloo (changes requested):

webapps/api/backend/views/vpools.py
```diff
@@ -251,3 +251,24 @@ def create_hprm_config_files(self, local_storagerouter, vpool, parameters):
         return VPoolController.create_hprm_config_files.delay(parameters=parameters,
                                                               vpool_guid=vpool.guid,
                                                               local_storagerouter_guid=local_storagerouter.guid)
+
+    @action()
+    @log()
+    @required_roles(['read', 'write'])
+    @return_task()
+    @load(VPool)
+    def move_multiple(self, vdisks, target_storagerouter_guid, force=False):
```
JeffreyDevloo: Change to: `def move_multiple(self, vpool, vdisk_guids, target_storagerouter_guid, force=False):`
The first parameter is initialized using the route param (/vpools/<vpool_guid>/move_multiple). The load decorator will then load a vPool object (the object type given to the load param) using the route param (in this case the vpool_guid)
All additional information comes in as a JSON serialized object with a key 'queryparams'. All keys within this queryparams represent the other params.
In our example, we would need to send:
```
{
    'queryparams': {'vdisk_guids': [<guid1>, ..],
                              'target_storagerouter_guid': <str_guid>}
}
```

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

PR #1932, opened 2018-01-17
bugfix vdisk creation gui

sjovdnbos:
Fixes a bug where invalid vdisk names were accepted in the gui, due to a lack of validation.

### DIFF
--- a/ovs/lib/vdisk.py
+++ b/ovs/lib/vdisk.py
@@ -841,8 +841,6 @@ class VDiskController(object):
         # Validations
         storagedriver = StorageDriver(storagedriver_guid)
         volume_name = str(volume_name)
-        if not re.match(VDisk.VDISK_NAME_REGEX, volume_name):
-            raise ValueError('Provided name did not match with the vDisk name regex')
         devicename = VDiskController.clean_devicename(volume_name)
         vpool = storagedriver.vpool
         if VDiskList.get_by_devicename_and_vpool(devicename, vpool) is not None:
--- a/ovs/lib/vdisk.py
+++ b/ovs/lib/vdisk.py
@@ -841,6 +841,7 @@ class VDiskController(object):
         # Validations
         storagedriver = StorageDriver(storagedriver_guid)
         devicename = VDiskController.clean_devicename(volume_name)
+        volume_name = str(volume_name)
         vpool = storagedriver.vpool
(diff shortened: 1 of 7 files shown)


ovs/lib/vdisk.py
```diff
@@ -841,6 +841,7 @@ def create_new(volume_name, volume_size, storagedriver_guid, pagecache_ratio=1.0
         # Validations
         storagedriver = StorageDriver(storagedriver_guid)
```
JeffreyDevloo, line 842: Add regex here

ovs/lib/vdisk.py
```diff
@@ -841,6 +841,7 @@ def create_new(volume_name, volume_size, storagedriver_guid, pagecache_ratio=1.0
...
         devicename = VDiskController.clean_devicename(volume_name)
```
JeffreyDevloo: Cast volume_name to a string here just be sure (could be a int due to decoding)

ovs/lib/vdisk.py
```diff
@@ -1588,7 +1589,7 @@ def clean_devicename(name):
         :return: A cleaned device name
         :rtype: str
         """
-        name = name.strip('/').replace(' ', '_')
+        name = str(name).strip('/').replace(' ', '_')
```
JeffreyDevloo: Clean device name expects a string, so better to check if the given item is a string rather than stringifying it yourself

JeffreyDevloo (changes requested):

ovs/lib/vdisk.py
```diff
@@ -1588,7 +1589,9 @@ def clean_devicename(name):
         :return: A cleaned device name
         :rtype: str
         """
-        name = name.strip('/').replace(' ', '_')
+        if not isinstance(name, str):
```
JeffreyDevloo: Handle unicodes (so instance(name, basestr)

ovs/lib/vdisk.py
```diff
@@ -1588,7 +1589,9 @@ def clean_devicename(name):
...
+            raise TypeError('Given name "{0}" could not be converted to a clean device name, str required'.format(name))
+        name = str(name).strip('/').replace(' ', '_')
```
JeffreyDevloo: Casting not needed as you are sure it is a string

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

PR #1970, opened 2018-02-09
statsmonkey restart masters

sjovdnbos:
Statsmonkey will now only shut down masters instead of all storagerouters


JeffreyDevloo (approved):

JeffreyDevloo:
Do not package this yet (Fix the other small tickets first)

PR #1971, opened 2018-02-12
Fix not all packages shown

sjovdnbos:


ovs/extensions/support/agent.py
```diff
@@ -195,16 +199,35 @@ def _set_client(self):
...
+
+        for thread in threads:
+            thread.join()
+
+        for _, d in versions_dict[self._client.ip].iteritems():
+            for package, version in d.iteritems():
+                version = str(version)
+                if package in final_dict:
+                    if version != final_dict[package]:
+                        pass #what to do
```
JeffreyDevloo: 2 spaces, capital case

JeffreyDevloo: I'd opt to keep the lowest value (the value should always be the same though)

JeffreyDevloo (changes requested):

ovs/extensions/support/agent.py
```diff
@@ -195,16 +199,35 @@ def _set_client(self):
...
+
+        for thread in threads:
+            thread.join()
+
+        for versions in versions_dict[self._client.ip].itervalues():
+            for package, version in versions.iteritems():
+                str_version = str(version)
+                if package in final_dict:
+                    if str_version != final_dict[package]:
+                        final_dict[package] = min(version, final_dict[package])
```
JeffreyDevloo: This is comparing strings. Might not give the result you want.
Instead do the followinging: build your dict as you are now but keep the looseversion item in it.
Before returning: covert all looseversions to strings

JeffreyDevloo (changes requested):

ovs/extensions/support/agent.py
JeffreyDevloo: Its a min(LooseVersion(), str)

ovs/extensions/support/agent.py
```diff
@@ -195,16 +200,42 @@ def _set_client(self):
...
+            thread.join()
+
+        for versions in versions_dict[self._client.ip].itervalues():
+            for package, version in versions.iteritems():
+                if package in final_dict:
+                    if version != final_dict[package]:
+                        final_dict[package] = min(version, final_dict[package])
+                else:
+                    final_dict[package] = version
+        sorted_dict = sorted(final_dict.items(), key=lambda t: t[0])
```
JeffreyDevloo: This is not a sorted dict but a list of tuples

JeffreyDevloo: Not necessary too (can be done in your dict comprehension instead)
OrderedDict((key, self._stringify_looseversion(value)) for key, value in sorted(dict.items(), key=lamba v: v[0])

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

PR #1994, opened 2018-03-07
Volume write buffer gui change

sjovdnbos:
Added additional parameters for creating a vpool volume write buffer 
-added volume write buffer and nondisposable sco factor as parameters, raising a warning upon subawesome parameter selection


JeffreyDevloo (changes requested):
Where is the API side of things?

Issue #2006, opened 2018-03-21
[Bighorn RC2] Support overview broken


JeffreyDevloo:
### Problem description
The package information and heartbeat data did not get displayed on the page

```
Mar 21 10:21:48 LC1SRV0008 celery[8683]: 2018-03-21 10:21:48 55100 +0100 - LC1SRV0008 - 8683/139926761338624 - celery/log.py - log - 98 - ERROR - Task ovs.storagerouter.get_support_metadata[7f299cf8-5ad4-4cb3-8357-68c9c341440c] raised unexpected: EncodeError(RepresenterError("cannot represent an object: OrderedDict([('ovs-albaproxy_myvpool01_0.service', 'loaded active running ALBA proxy'), ('ovs-albaproxy_myvpool01_1.service', 'loaded active running ALBA proxy'), ('ovs-dtl_myvpool01.service', 'loaded active running Open vStorage distributed transaction log'), ('ovs-scheduled-tasks.service', 'loaded active running Open vStorage scheduled tasks'), ('ovs-support-agent.service', 'loaded active running Open vStorage support agent'), ('ovs-volumedriver_myvpool01.service', 'loaded active running Open vStorage StorageDriver'), ('ovs-volumerouter-consumer.service', 'loaded active running Open vStorage volumerouter consumer'), ('ovs-watcher-config.service', 'loaded active running Watcher for the configuration cluster'), ('ovs-watcher-framework.service', 'loaded active running Watcher for Open vStorage master services'), ('ovs-watcher-volumedriver.service', 'loaded active running Watcher for Open vStorage master services'), ('ovs-webapp-api.service', 'loaded active running Open vStorage webapp API'), ('ovs-workers.service', 'loaded active running Open vStorage workers')])",),)
Mar 21 10:21:48 LC1SRV0008 celery[8683]: Traceback (most recent call last):
Mar 21 10:21:48 LC1SRV0008 celery[8683]:   File "/usr/lib/python2.7/dist-packages/celery/app/trace.py", line 283, in trace_task
Mar 21 10:21:48 LC1SRV0008 celery[8683]:     uuid, retval, SUCCESS, request=task_request,
Mar 21 10:21:48 LC1SRV0008 celery[8683]:   File "/usr/lib/python2.7/dist-packages/celery/backends/base.py", line 271, in store_result
Mar 21 10:21:48 LC1SRV0008 celery[8683]:     request=request, **kwargs)
[...]

PR #2007, opened 2018-03-21
Support OrderedDict items to be stored in ArakoonResult

JeffreyDevloo:
Addresses https://github.com/openvstorage/framework/issues/2006


JeffreyDevloo:
Re-opening as base master...

PR #2015, opened 2018-03-28
Node id extensions dep

sjovdnbos:
openvstorage_id is now assigned in the extensions package installation


JeffreyDevloo (changes requested):
I wouldn't remove it just yet. Keeping this is in would prevent release failures (the ID won't be generated if the right ovs_extensions package is installed anyway)

Issue #2027, opened 2018-04-18
Allow longer vdisk names


JeffreyDevloo:
Team India is bumping into the vdisks 50 char constraint on Bighorn

JeffreyDevloo:
Fixed by #2030 
Packaged in openvstorage-core_2.12.0-dev.1524232601.13075b8-1_amd64.deb

JeffreyDevloo:
Released in https://github.com/openvstorage/framework/releases/tag/2.12.0
Packaged in openvstorage-core_2.12.0-1_amd64.deb

PR #2030, opened 2018-04-20
Allow vdisk names to be longer

JeffreyDevloo:
Addresses https://github.com/openvstorage/framework/issues/2027

### DIFF
--- a/ovs/dal/hybrids/vdisk.py
+++ b/ovs/dal/hybrids/vdisk.py
@@ -40,7 +40,7 @@ class VDisk(DataObject):
     """
     STATUSES = DataObject.enumerator('Status', ['HALTED', 'NON_RUNNING', 'RUNNING', 'UNKNOWN'])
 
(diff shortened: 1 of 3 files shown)


sjovdnbos (approved):

ovs/dal/hybrids/vdisk.py
```diff
@@ -40,7 +40,7 @@ class VDisk(DataObject):
     """
     STATUSES = DataObject.enumerator('Status', ['HALTED', 'NON_RUNNING', 'RUNNING', 'UNKNOWN'])
 
-    VDISK_NAME_REGEX = '^[0-9a-zA-Z][\-_a-zA-Z0-9]{1,48}[a-zA-Z0-9]$'
+    VDISK_NAME_REGEX = '^[0-9a-zA-Z][\-_a-zA-Z0-9]{1,98}[a-zA-Z0-9]$'
```
toolslive: why limit this at all?

JeffreyDevloo: The character limitation was put in place by the previous guys but never enforced due to a bug.
So as to why it should be limited... Force the customer to think about their names...? :p

PR #2110, opened 2018-06-28
scrubber stack deployment fix

sjovdnbos:
Fixed an issue where the scrubber did not result in logging errors when it failed on unexpected moments


JeffreyDevloo (changes requested):
This will only capture errors from the StackWorker init.
The deploy_stack_and_scrub can still error without the errors being logged.

PR #2115, opened 2018-07-04
Introduce graphite statistics

sjovdnbos:
The framework can currently cope with scrubber data to be send and in a next stage also healthcheck nsm data.


ovs/extensions/generic/graphiteclient.py
```diff
@@ -0,0 +1,76 @@
...
+        :param port: port of the UDP listening socket
+        :type port: int
+        """
+        config_path = 'ovs/framework/graphite'
+        self.precursor = 'openvstorage.fwk.{0} {1} {2}'
+
+        try:
+            graphite_data = Configuration.get(config_path)
+        except NotFoundException:
+            raise RuntimeError('No graphite data found in config path `{0}`'.format(config_path))
```
JeffreyDevloo: Shouldn't the raise only happen when no ip/port were specified?

ovs/extensions/generic/graphiteclient.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def send(self, path, data):
+        # type: (str, float) -> None
+        """
+        Send the statistics with client
+        :param path: path in graphite to send the data to
+        :param data: data to send
+        :return: None
+        """
+        sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
```
JeffreyDevloo: Shouldn't you fire and forget here? A try except for all socket errors perhaps?

ovs/lib/graphite.py
```diff
@@ -0,0 +1,64 @@
...
+#
+# Open vStorage is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY of any kind.
+
+import time
+from ovs.extensions.generic.graphiteclient import GraphiteClient
+from ovs.extensions.generic.configuration import Configuration
+
+
+class GraphiteController(GraphiteClient):
```
JeffreyDevloo: Not seeing the added value of this class
It inherits from GraphiteClient but it never initializes itself. It also build a new client each time?
Just remove this client and add the methods within the scrubber itself...

webapps/frontend/app/viewmodels/services/storagerouter.js
```diff
@@ -47,7 +46,34 @@ define([
             return api.post('storagerouters/' + storageRouterGuid + '/get_metadata')
                 .then(shared.tasks.wait)
         };
-
+        /**
+         * Fetches support info of a StorageRouter
+         * @param storageRouterGuid: Guid of the StorageRouter
+         * @return {Promise<T>}
+         */
+        self.getSupportSettingsHandle = function(storageRouterGuid) {
```
JeffreyDevloo: Handles are used to limit api calls. This should be getSupportInfo instead.
The handles is the XHRRequest that the api.get returns. It is then used to check if the xhr was already resolved or not before re-querying.

webapps/frontend/app/viewmodels/site/admin/support.js
```diff
@@ -153,77 +157,83 @@ define([
...
                                 storageRouter.fillData(srData[storageRouter.guid()]);
                             }
                             storageRouter.loading(true);
                             var calls = [];
                             if (index === 0) {  // Support information are cluster-wide settings, so only retrieving for 1st StorageRouter
                                 if (generic.xhrCompleted(self.supportSettingsHandle)) {
-                                    self.supportSettingsHandle = api.get('storagerouters/' + storageRouter.guid() + '/get_support_info')
-                                        .then(self.shared.tasks.wait)
-                                        .then(function(data) {
+                                    self.supportSettingsHandle = storagerouterService.getSupportSettingsHandle(storageRouter.guid())
```
JeffreyDevloo: The api call is not a Handle

webapps/frontend/app/viewmodels/site/admin/support.js
```diff
@@ -153,77 +157,83 @@ define([
...
+                                        .then(function (data) {
+                                            console.log(data)
```
JeffreyDevloo: Remove log

JeffreyDevloo (changes requested):

ovs/extensions/generic/graphiteclient.py
sjovdnbos: If no graphite  configuration data is found -> no ip/portdata is passed, and an error is raised due to not enough info. Is this not expected?

ovs/extensions/generic/graphiteclient.py
sjovdnbos: a socket is needed to use UDP with python, as far as i know? any pointers to other info?

ovs/lib/graphite.py
sjovdnbos: The controller should bundle all calls, where the client does the connection work, and doesnt care what exactly he sends towards graphite. 
The graphiteclient beind rebuild everytime is bad though, indeed

ovs/extensions/generic/graphiteclient.py
JeffreyDevloo: You can pass a ip port yourself. You should not care for the graphite data in that case

ovs/extensions/generic/graphiteclient.py
JeffreyDevloo: Do errors get raised when the host is not available? I'd test these corner cases and see if try-except is required.

ovs/lib/graphite.py
JeffreyDevloo: The controller should not inherit from the client but should be instantiated so it can make the graphite client for further use.
```
class GraphiteController(object):
    def __init__(self):
         self._graphite_client = GraphiteClient()
   // Scrubber stats
   def send...
```

ovs/extensions/generic/graphiteclient.py
sjovdnbos: Only constraint seems to be the format of the parameters. The host does not need to be accessible. As the passed datastring is already ensured to be a string, no errors need to be caught

ovs/extensions/generic/graphiteclient.py
```diff
@@ -0,0 +1,76 @@
...
+        datastring = self.precursor.format(path, int(data), int(time.time()))  # Carbon timestamp in integers
+        sock.sendto(datastring, (self.ip, self.port))
+        sock.close()
```
JeffreyDevloo: Wrap in a try finally. finally should close the socket

ovs/lib/helpers/generic/scrubber.py
```diff
@@ -620,13 +620,17 @@ def _verify_mds_config(current_vdisk):
                             # Do the actual scrubbing
                             with vdisk.storagedriver_client.make_locked_client(str(vdisk.volume_id), update_interval_secs=lease_interval) as locked_client:
                                 self._logger.info('{0} - Retrieve and apply scrub work'.format(vdisk_log, vdisk.name))
+                                starttime = time.time()
+                                graphite_controller = GraphiteController()
```
JeffreyDevloo: A single controller for the whole scrub job should be enough.
Declare it in the class Scrubber and pass it along to all threads

ovs/lib/storagerouter.py
```diff
@@ -364,6 +376,17 @@ def configure_support(support_info):
                         StorageRouterController._service_manager.stop_service(name=service_name, client=root_client)
                         StorageRouterController._service_manager.remove_service(name=service_name, client=root_client)
 
+        # Configure scrubbing statistics
```
JeffreyDevloo: The graphite endpoint is configured here. Nothing scrubber related

webapps/frontend/app/viewmodels/services/storagerouter.js
```diff
@@ -47,7 +46,34 @@ define([
...
+            return api.get('storagerouters/' + storageRouterGuid + '/get_support_info')
+                .then(shared.tasks.wait)
+        };
+        /**
+         * Fetches support metadata of a StorageRouter
+         * @param storageRouterGuid: Guid of the StorageRouter
+         * @return {Promise<T>}
+         */
+        self.getSupportMetadataHandle = function(storageRouterGuid) {
```
JeffreyDevloo: Not an actual Handle. See feedback of the other handle

webapps/frontend/app/viewmodels/site/admin/support.js
```diff
@@ -109,23 +121,59 @@ define([
...
+                    }
+                }
+            });
+            wizard.finishing.always(function() {
+                if (self.supportSettingsChanged() === true) {
+                    self.save();
+                }
+            });
+            dialog.show(wizard);
+            return true;  // Critical to return True, otherwise knockout can't handle a click + checked event
```
JeffreyDevloo: Returning True allows your browser to run the default click behavior. This comment makes it look like knockout is incompetent at it

webapps/frontend/app/viewmodels/site/admin/support.js
```diff
@@ -44,13 +52,14 @@ define([
 
         // Handles
         self.supportMetadataHandle = {};
-        self.supportSettingsHandle = undefined;
+        self.supportSettings = undefined;
```
JeffreyDevloo: Renaming might cause confusion. This is a handle to avoid multiple concurrent requests.

webapps/frontend/app/viewmodels/services/storagerouter.js
```diff
@@ -47,7 +46,34 @@ define([
...
+            return api.get('storagerouters/' + storageRouterGuid + '/get_support_metadata')
+                .then(shared.tasks.wait)
+        };
+        /**
+         * Fetches support info of a StorageRouter
```
JeffreyDevloo: It sets new support info

webapps/frontend/app/viewmodels/services/storagerouter.js
```diff
@@ -47,7 +46,34 @@ define([
...
+         * @param storageRouterGuid: Guid of the StorageRouter
+         * @param metaData: data to post
+         * @return {Promise<T>}
+         */
+        self.postSupportData = function(storageRouterGuid, metaData) {
```
JeffreyDevloo: I'd call it saveSupportdata.
Also it is not metadata but data to. Rename the parameter to reflect that.

webapps/frontend/app/viewmodels/wizards/graphiteconfigure/data.js
```diff
@@ -0,0 +1,24 @@
...
+// as published by the Free Software Foundation, in version 3 as it comes
+// in the LICENSE.txt file of the Open vStorage OSE distribution.
+//
+// Open vStorage is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY of any kind.
+/*global define */
+define([
+], function(){
+    "use strict";
+    var singleton = function() {
```
JeffreyDevloo: Use the new type of data setting (with stepOPtions in the index) to avoid cleaning the options and multiple instances mingling with eachother

webapps/frontend/app/viewmodels/services/statistics.js
```diff
@@ -0,0 +1,47 @@
...
+         */
+        self.graphiteStatistics = function() {
+            return api.get('ovs/statistics/graphite');
+        };
+        /**
+         * Generates an empty slot for an AlbaNode
+         * @param metaData: host IP and port
+         * @return {*|Promise}
+         */
+        self.postStatistics = function(metaData) {
```
JeffreyDevloo: Call it saveStatistics and rename metadata to data. It's data you post, not data about data :)

webapps/frontend/app/viewmodels/services/statistics.js
```diff
@@ -0,0 +1,47 @@
...
+        var self = this;
+
+
+        // Functions
+        /**
+         * Fetch the currently saved graphite credentials
+         * @returns {Deferred}
+         */
+        self.graphiteStatistics = function() {
+            return api.get('ovs/statistics/graphite');
```
JeffreyDevloo: This path is not right. Looks like a configuration path instead of an API one?
Edit: revisiting this file because all saving is done using the postSupportSettings so this service is not used at all?

JeffreyDevloo (changes requested):

ovs/extensions/generic/graphiteclient.py
```diff
@@ -0,0 +1,117 @@
...
+
+import time
+import socket
+from ovs.extensions.generic.configuration import Configuration
+from ovs_extensions.generic.configuration.exceptions import ConfigurationNotFoundException as NotFoundException
+from ovs_extensions.generic.toolbox import ExtensionsToolbox
+from ovs.extensions.generic.logger import Logger
+
+
+class GraphiteClient(object):
```
JeffreyDevloo: Make a base GraphiteClient in the extensions. We might need it in different projects too

ovs/lib/graphite.py
```diff
@@ -0,0 +1,83 @@
...
+
+class GraphiteController():
+    """
+    Graphite Controller that sends or does not send, depending on the saved config setting
+    """
+
+    def __init__(self, database=None):
+        celery_scheduling = Configuration.get(key='/ovs/framework/scheduling/celery', default={})
+        self._send_statistics = any(celery_scheduling.get(key) is not None for key in ['ovs.stats_monkey.run_all', 'alba.stats_monkey.run_all'])
+        if self._send_statistics:
```
JeffreyDevloo: Set self._client = None here too. If self._send_statistics would be False, the class would have one property less otherwise.

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

PR #2124, opened 2018-07-16
added unittest for httprequest properwrap

sjovdnbos:


ovs/lib/tests/generic_tests/test_decorators.py
```diff
@@ -0,0 +1,44 @@
...
+
+        @wrap(key_wrapper)
+        def test_wrapper_decorator():
+            return value_wrapper
+
+        wrapped_dict = test_wrapper_decorator()
+        self.assertDictEqual(d1=wrapped_dict, d2= {key_wrapper: value_wrapper})
+
+        @wrap(key_wrapper)
+        def test_wrapper_decorator(a='a'):
```
JeffreyDevloo: This test is obsolete. It does not test anything extra. If test 1 would pass, this one would also always pass.

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

PR #2139, opened 2018-08-07
implement the ClusterNotReachable exception 

sjovdnbos:
Replace occurrences of Runtime error with string checking by volumedriver ClusterNotReachable

### DIFF
--- a/ovs/extensions/storageserver/storagedriver.py
+++ b/ovs/extensions/storageserver/storagedriver.py
@@ -441,5 +441,4 @@ def is_connection_failure(exception):
     :type exception: Exception
     :return: bool
     """
-    return isinstance(exception, ClusterNotReachableException) or (isinstance(exception, RuntimeError) and 'failed to send XMLRPC request' in str(exception)):
-
+    return isinstance(exception, ClusterNotReachableException) or (isinstance(exception, RuntimeError) and 'failed to send XMLRPC request' in str(exception))
--- a/ovs/extensions/storageserver/storagedriver.py
+++ b/ovs/extensions/storageserver/storagedriver.py
@@ -31,7 +31,7 @@ from volumedriver.storagerouter import storagerouterclient
 # noinspection PyUnresolvedReferences
 
 from volumedriver.storagerouter.storagerouterclient import \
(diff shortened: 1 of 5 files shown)


ovs/extensions/storageserver/storagedriver.py
```diff
@@ -371,9 +371,9 @@ def save(self, client=None, force_reload=False):
                 with remote(client.ip, [LocalStorageRouterClient]) as rem:
                     changes = copy.deepcopy(rem.LocalStorageRouterClient(self.remote_path).update_configuration(self.remote_path))
             reloaded = True
-        except Exception as ex:
-            if 'ClusterNotReachableException' not in str(ex):
-                raise
+        except (ClusterNotReachableException, RuntimeError) as exception:
```
JeffreyDevloo: This extra checking is done quite often in this PR.
Isn't it better to let a function decide if it was a ClusterNotReachble/RuntimeError with the message?

JeffreyDevloo (changes requested):

ovs/extensions/storageserver/storagedriver.py
```diff
@@ -433,3 +433,9 @@ def _delete(self, section):
         """
         if section in self.configuration:
             del self.configuration[section]
+
+def is_clusterNotReachableException(exception):
```
JeffreyDevloo: This does not look right

ovs/extensions/storageserver/storagedriver.py
```diff
@@ -433,3 +433,9 @@ def _delete(self, section):
...
+    if isinstance(exception, ClusterNotReachableException) or (isinstance(exception, RuntimeError) and 'failed to send XMLRPC request' in str(exception)):
+        return True
```
JeffreyDevloo: `return isinstance(exception, ClusterNotReachableException) or (isinstance(exception, RuntimeError) and 'failed to send XMLRPC request' in str(exception)):` already provide the True - False. No need to check it again

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

PR #2177, opened 2018-08-31
Lower reaction time of some input fields

sjovdnbos:
These two number input fields were affected by a change that bumped input reactiontimes for a more intuitive GUI. These two changes did not affect the GUI in a bad way though, and reaction time is now lowered.

### DIFF
--- a/webapps/frontend/app/viewmodels/containers/storagedriver/configuration.js
+++ b/webapps/frontend/app/viewmodels/containers/storagedriver/configuration.js
@@ -27,7 +27,7 @@ define([
     var ConfigurationViewModel = function(data) {
         var self = this;
         self.storageIP = ko.observable('').extend({regex: generic.ipRegex});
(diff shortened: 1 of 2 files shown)


webapps/frontend/app/viewmodels/containers/storagedriver/configuration.js
```diff
@@ -27,7 +27,7 @@ define([
     var ConfigurationViewModel = function(data) {
         var self = this;
         self.storageIP = ko.observable('').extend({regex: generic.ipRegex});
-        self.proxyAmount = ko.observable(2).extend({numeric: {min: 1, max: 16}, rateLimit: { method: "notifyWhenChangesStop", timeout: 800 }});
+        self.proxyAmount = ko.observable(2).extend({numeric: {min: 1, max: 16}, rateLimit: { method: "notifyWhenChangesStop", timeout: 1 }});
```
JeffreyDevloo: Why the 1ms timeout? You could remove it

JeffreyDevloo (changes requested):

JeffreyDevloo (approved):

PR #2178, opened 2018-09-03
S3 transaction arakoon type

JeffreyDevloo:


ovs/extensions/plugins/apiclient.py
```diff
@@ -86,8 +104,9 @@ def _call(self, method, url, data=None, timeout=None, clean=False):
                   'headers': self._base_headers,
                   'verify': False,
                   'timeout': timeout}
-        if data is not None:
-            kwargs['data'] = data
+        for key, val in [('json', json), ('data', data)]:
```
sjovdnbos: Could you provide some comments on why you do this?

ovs/extensions/plugins/apiclient.py
```diff
@@ -121,7 +140,7 @@ def _refresh(self):
         return self._base_url, self._base_headers
 
     def extract_data(self, response_data, old_key=None):
-        # type: (dict) -> any
+        # type: (dict, str) -> any
```
sjovdnbos: Optional[str]

sjovdnbos (changes requested):

ovs/extensions/plugins/apiclient.py
JeffreyDevloo: Requests library can both take in 'data' or 'json' keyword.
When data is given: no extra heading is added and the data is serialized as json
When json is given: the 'Content type: Application/json' header is added.
The loop is to provide both options

sjovdnbos (approved):

PR #2219, opened 2018-11-23
Extend volumedriver config objects to detect which option is dynamically reloadable

JeffreyDevloo:


ovs/extensions/storageserver/storagedriverconfig/base.py
```diff
@@ -0,0 +1,203 @@
...
+        # type: () -> List[VolumeDriverConfigOption]
+        """
+        Get all possible options
+        :return: A list with all options
+        :rtype: List[VolumeDriverConfigOption]
+        """
+        if not cls._options_by_component_and_key:
+            cls._cache_options()
+        options = []
+        for component, component_keys_options in cls._options_by_component_and_key.iteritems():
```
JeffreyDevloo: Not sure if working with a dict is beneficial on this small set of items. It'd be more clear to use a list and filter on it instead

JeffreyDevloo:
Closing because it's insufficient atm

ovs/extensions/storageserver/storagedriverconfig/base.py
sjovdnbos: Agreeing on this

ovs/extensions/storageserver/storagedriverconfig/base.py
```diff
@@ -0,0 +1,203 @@
...
+        Retrieve an option by component and key
+        Returns None when no option was found
+        :param component: Component to retrieve
+        :type component: str
+        :param key: Key to retrieve from component
+        :type key: str
+        :return: Option if found else none
+        :rtype: VolumeDriverConfigOption
+        """
+        if not cls._options_by_component_and_key:
```
sjovdnbos: Instead of checking every time, wouldn't always calling a function `_validate_cache` or sth and differentiating in behaviour there, be more readable?

ovs/extensions/storageserver/storagedriverconfig/base.py
```diff
@@ -0,0 +1,213 @@
...
+
+class VolumeDriverConfigOption(object):
+    """
+    Config option read out by 'volumedriver_fs', '--config-help-markdown'
+    """
+    _options = []
+
+    def __init__(self, component, key, default_value, dynamically_reconfigurable, remarks):
+        #  type: (str, str, any, bool, str) -> None
+        """
```
sjovdnbos, line 44: You never check typings of the inputs. This is fine as it is, but in a second phase, where these options will be configurable in the GUI, this will need to be implemented in some place. And i propose to do that.. here?

ovs/extensions/storageserver/storagedriverconfig/storagedriver.py
```diff
@@ -39,7 +41,9 @@ def __init__(self, vregistry_arakoon_cluster_id, vregistry_arakoon_cluster_nodes
         self.vregistry_arakoon_timeout_ms = vregistry_arakoon_timeout_ms
 
 
-class DtlConfig(GenericConfig):
+class DistributedTransactionLogConfig(BaseStorageDriverConfig):
+
+    component_identifier = 'distributed_transaction_log'
```
sjovdnbos, line 46: do these strings have no `constants` files to put them in?

sjovdnbos: edit after some more lines of reviewing: please offload to `constants` file

sjovdnbos (changes requested):

ovs/extensions/storageserver/storagedriverconfig/storagedriver.py
JeffreyDevloo: They are only written once.

ovs/extensions/storageserver/storagedriverconfig/base.py
JeffreyDevloo: This isn't the good spot. The good spot will be BaseStorageDriverConfig and the validation will be done by the volumedriver itself. Nothing to do here

ovs/extensions/storageserver/storagedriverconfig/base.py
```diff
@@ -0,0 +1,213 @@
...
+        Returns None when no option was found
+        :param component: Component to retrieve
+        :type component: str
+        :param key: Key to retrieve from component
+        :type key: str
+        :return: Option if found else none
+        :rtype: VolumeDriverConfigOption
+        """
+        try:
+            return filter(lambda option: option.key == key and option.component == component, cls._options)[0]
```
sjovdnbos: This will always check an entire list and then just take the first value of the obtained list.
Computationally, it would be more efficient to use a for loop and break out of the loop upon finding a value.
Aesthetically, it would be cleaner to use a list comprehension. 
Wouldn't either of this be a better fit here?

ovs/extensions/storageserver/storagedriverconfig/base.py
```diff
@@ -0,0 +1,213 @@
...
+        Retrieve all options for a component
+        :param component: Component to retrieve options for
+        :type component: str
+        :return: All options associated with the component
+        :rtype: List[VolumeDriverConfigOption]
+        """
+        return filter(lambda option: option.component == component, cls._options)
+
+    @classmethod
+    def _cache_options(cls):
```
sjovdnbos: This is not an ideal function name. 
`_read_voldr_config` or sth would be better imho

ovs/extensions/storageserver/storagedriverconfig/base.py
```diff
@@ -0,0 +1,213 @@
...
+        :rtype: NoneType
+        """
+        options = []
+        markdown = check_output(['volumedriver_fs', '--config-help-markdown'])
+        for line in markdown.split('\n|')[2:]:
+            options.append(cls.from_option_line('|' + line))
+        cls._options = options
+
+    @classmethod
+    def from_option_line(cls, line):
```
sjovdnbos: Not an ideal function name either tbh
`parse_line`, `read_line`, use a verb to show what it does

ovs/extensions/storageserver/storagedriverconfig/base.py
```diff
@@ -0,0 +1,213 @@
...
+        Check if the other value is not equal to this instance
+        :param other: Other value to compare to
+        :type other: BaseStorageDriverConfig
+        :return: True if not equal else false
+        :rtype: bool
+        """
+        return not other == self
+
+    def get_difference(self, other):
+        # type: (BaseStorageDriverConfig) -> Dict[str, any]
```
sjovdnbos: You output a list, not a dict

sjovdnbos (changes requested):

sjovdnbos (approved):

PR #2280, opened 2019-03-19
Implement the storagedriver dynamic 'status'.

sjovdnbos:


ovs/dal/hybrids/storagedriver.py
```diff
@@ -62,8 +65,16 @@ def _status(self):
         """
         Fetches the Status of the Storage Driver.
         """
-        _ = self
-        return None
+        vpool_status = self.vpool.status
```
JeffreyDevloo: This is not right, right?
The vpool can be installing a different storagedriver

ovs/lib/vpool.py
```diff
@@ -139,6 +139,7 @@ def add_vpool(cls, parameters):
             # MODEL STORAGEDRIVER AND PARTITION JUNCTIONS
             sd_installer.create()
             sd_installer.create_partitions()
+            sd_installer.storagedriver.save()
```
JeffreyDevloo: Why save here?

JeffreyDevloo (changes requested):

ovs/lib/vpool.py
sjovdnbos: Relic from first commits. woops

JeffreyDevloo (approved):

PR #2332, opened 2019-08-01
[wip] discard celery related changes in the volumedriver update

sjovdnbos:
Revert celery related changes that make the volumedriver update loop forever.

### DIFF
--- a/ovs/update/volumedriver/updater.py
+++ b/ovs/update/volumedriver/updater.py
@@ -87,9 +87,6 @@ class VolumeDriverUpdater(ComponentUpdater):
                 # Get the migration plans for every volume on this host. If there are no plans for certain volumes, it will raise
                 balances_by_vpool = cls.get_vpool_balances_for_evacuating_storagerouter(cls.LOCAL_SR)
                 if initial_run_steps:
-                    cls.logger.info('Offloading a MDS catchup to celery. This will ensure all slaves will be caught up to avoid deadlocking')
-                    MDSServiceController.mds_catchup.apply_async()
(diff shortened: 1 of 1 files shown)


ovs/update/volumedriver/updater.py
```diff
@@ -87,9 +87,6 @@ def restart_services(cls):
                 # Get the migration plans for every volume on this host. If there are no plans for certain volumes, it will raise
                 balances_by_vpool = cls.get_vpool_balances_for_evacuating_storagerouter(cls.LOCAL_SR)
                 if initial_run_steps:
-                    cls.logger.info('Offloading a MDS catchup to celery. This will ensure all slaves will be caught up to avoid deadlocking')
-                    MDSServiceController.mds_catchup.apply_async()
```
JeffreyDevloo, line 91: Has to be kept. We dont care about the result, it just runs in the background

ovs/update/volumedriver/updater.py
```diff
@@ -226,26 +213,20 @@ def migrate_master_mds(cls, storagerouter, max_chain_size=100, group_timeout=10
         """
         cls.logger.info("Starting MDS migrations")
         while True:
-            hosted_vdisk_guids = storagerouter._vdisks_guids()
```
JeffreyDevloo, line 229: Dont use signatures here as it will lock up. Discarded tasks are still being waited for

The reason for the catchup (see comment above) is to ensure all slaves are caught up with their master

JeffreyDevloo (changes requested):

